PR #35, opened 2026-04-11
12 simple main menu

### LINKED ISSUE #12: Simple Main Menu
- as simple as possible

51th08:


include/ui/renderer.hpp
```diff
@@ -1,20 +1,28 @@
 #pragma once
 
 #include "controller/view/view.hpp"
+#include <SFML/Graphics/Font.hpp>
```
5tefan543, line 4: Can we include this in view.hpp to reduce imports here?

include/ui/renderer.hpp
```diff
@@ -1,20 +1,28 @@
...
 #include <SFML/Graphics/RenderWindow.hpp>
+#include <SFML/Graphics/Text.hpp>
```
5tefan543, line 6: Can we include this in view.hpp to reduce imports here?

include/ui/ui.hpp
```diff
@@ -23,6 +25,7 @@ class UI {
     bool isOpen() const;
     controller::InputState pollInput();
     void render(const controller::View &view);
+    sf::Font getFont(const controller::Font font);
```
5tefan543, line 28: Is this needed in ui.hpp? It is not implemented.

src/controller/state/state.cpp
```diff
@@ -1,4 +1,5 @@
 #include "controller/state/state.hpp"
+#include "controller/view/text.hpp"
```
5tefan543, line 2: Include text.hpp in view.hpp then we don't need this here

src/ui/renderer.cpp
```diff
@@ -58,6 +72,27 @@ void Renderer::renderItem(sf::RenderWindow &window, const controller::Button &bu
     rect.setPosition({centerX, centerY});
 
     window.draw(rect);
+
+    sf::Text text = toSfText(button.text);
+    // Calc button text position based on button's position
```
5tefan543, line 77: Could we not call renderItems here since Text is a ViewItem

5tefan543 (changes requested):

src/ui/renderer.cpp
51th08: That was initial intention but the texts position depends on the button which is not available with just a renterItem() call.

renderItem uses - as all other render functions - no absolute coords but a center offset depdent on the window size. Positioning the text on top of the button would therefore be only possible if the text's offset would be updated with e.g. `- (window.getSize().x / 2.0f)` before the renderItem call. I find the current solution better than this hack. It would introduce specific logic dependencies/assumptions to the text's render method.

Do you have some specific fix in mind?
Maybe change the way positions are saved? E.g. simple enum in viewItem dictating if (x,y) is absolute, relative. The text's position could be set to an absolute position here before calling its renderItem() which handles absolute/relative accordingly. 
-> This could also pave the way for dimensions relative to window size.

src/controller/state/state.cpp
51th08: Is that a common convention?

view.hpp does not use anything from text.hpp directly. Other parts like main.cpp don't require text.hpp but have to include view.hpp. I'm used to import only what is required where it is required.

src/ui/renderer.cpp
5tefan543: Yes, you’re right, handling the button-specific text directly within the button rendering logic is cleaner this way.

5tefan543 (changes requested):

include/ui/renderer.hpp
5tefan543: Ah, I did not see that these were SFML includes. Sorry :D

include/ui/ui.hpp
5tefan543: Ah, I did not see that these were SFML includes. Sorry :D

include/ui/renderer.hpp
5tefan543: Ah, I did not see that these were SFML includes. Sorry :D

src/controller/state/state.cpp
5tefan543: Okay, you've convinced me :)

5tefan543 (approved):

PR #40, opened 2026-04-12
Integrate ImGui as a debug UI

### LINKED ISSUE #18: Integrate Imgui as debug ui
- integrate dependency
- create example dialog in some state (e.g. game or menu)

5tefan543:


src/ui/ui.cpp
```diff
@@ -10,36 +11,57 @@ namespace ui {
...
 bool UI::isOpen() const
 {
     return window.isOpen();
 }
 
 controller::InputState UI::pollInput()
 {
-    return input_handler.pollInput(window);
+    inputState = inputHandler.pollInput(window);
+    return inputState;
```
RZacchia, line 46: Is there a reason for not in-lining the return value?

tests/controller/state/state_manager_test.cpp
RZacchia, line 1: Some Tests have compile errors because they stil use the old update signature
state_test.cpp
game_test.cpp

src/ui/ui.cpp
5tefan543: Yes `inputState` is a member var of `UI` since user control is needed in DebugUI

Maybe we should start to use some convention for class member variables?

Here a response from ChatGPT

---

# C++ Naming Conventions for Class Member Variables

## Recommended Convention (Modern C++)

### Trailing underscore for members

```cpp
class Player {
private:
    int health_;
    float speed_;
    std::string name_;
};
```

### Why this is popular:
- Clean and readable
- Avoids naming conflicts with parameters
- Widely used in modern C++ (e.g. Google style)

```cpp
Player(std::string name, int health)
    : name_(std::move(name)), health_(health) {}
```

---

## Alternative: Prefix with `m_` (older style)

```cpp
class Player {
private:
    int m_health;
    float m_speed;
};
```

### Pros:
- Explicit that it's a member variable

### Cons:
- More verbose
- Considered outdated in modern C++

---

## Minimalist Style (not recommended)

```cpp
class Player {
private:
    int health;
};
```

### Issue:
Requires `this->` to disambiguate:

```cpp
Player(int health) {
    this->health = health;
}
```

---

## Less Common Variations

### Double underscore (avoid)
```cpp
int health__;
```

### Leading underscore (avoid)
```cpp
int _health;
```

⚠️ These can conflict with reserved identifiers in C++

---

## Recommendation

Use:

```cpp
int health_;
```

### Why:
- Modern standard
- Clean constructors
- Avoids `this->`
- Works well in larger codebases

---

## Rule of Thumb

| Style        | Use Case                          |
|--------------|----------------------------------|
| `health_`    | ✅ Modern C++ (recommended)       |
| `m_health`   | Legacy / enterprise              |
| `health`     | Small/simple codebases           |

---

## Conclusion

Pick one convention and stay consistent across your entire project.

5tefan543: -> handled in issue #42

tests/controller/state/state_manager_test.cpp
5tefan543: Now fixed

RZacchia (approved):

PR #41, opened 2026-04-15
Fix unused parameter warnings

51th08:
Fix unused parameter and return type warnings.

They are annoying :)

### DIFF
--- a/src/controller/state/state.cpp
+++ b/src/controller/state/state.cpp
@@ -10,7 +10,7 @@ std::unique_ptr<MenuState> MenuState::createMenu(MenuType menuType)
     return menu;
 }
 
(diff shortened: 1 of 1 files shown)


src/controller/state/state.cpp
```diff
@@ -158,6 +158,8 @@ std::string MenuState::toString() const
         return "PauseMenu";
     case MenuType::GameOverMenu:
         return "GameOverMenu";
+    default:
+        return "Unknown (prob. forgot to implement) MenuState";
```
5tefan543, line 162: remove `(prob. forgot to implement)`
and add test case for default branch

5tefan543 (changes requested):

5tefan543:
-> handled in PR #40

PR #45, opened 2026-04-19
Refactor View ownership in State

51th08:
Summary:

- Add trailing _ for protected members (clang-tidy)
- All states now have a view_ member (inherited from BaseState)
- Menu Buttons are now handled via shared pointers to save a reference in dedicated Button list (where index = Button ID for selection identification)


51th08:
Please look at the tests. No Idea if this solution is the proper catch2 way to test multiple state (menuState) transitions in a linear fashion. The linear testing requirement comes from testing the button selection action (e.g. left/right btn) implicitly through the stateTransition return type after a confirmation input.

include/controller/state/state.hpp
```diff
@@ -17,22 +17,30 @@ struct BaseState {
...
-    static std::unique_ptr<MenuState> createMenu(MenuType menuType);
+    static std::unique_ptr<MenuState> createMenu(const MenuType menuType);
 
     StateTransitionAction update(const InputState &input, DebugContext &debug, float dt) override;
-    View getView() override;
     std::string toString() const override;
+
+  private:
+    std::size_t selectedButtonID_ = 0;
+    std::vector<std::shared_ptr<Button>> buttons_; // Shares Button with view_.items
```
5tefan543: I don't really like this design with the shared ptr. I understand that you don't want to iterate over the ViewItems of the card each time (even though this is not a performance problem and rather only verbose code) but I don't think that justifies shared ownership here.

In my opinion, a (Menu)State should own all elements (like the Buttons) where it changes their data. However, the view only needs access to the button and therefore should not own it.

Therefore, I would propose to simply store the buttons in the MenuState and store a pointer in the View.

Even nicer would be to store a reference in the view since that is definitely a signal that the View does not own the buttons. However, with a simple reference this is not possible.

But there exists [std::reference_wrapper](https://en.cppreference.com/cpp/utility/functional/reference_wrapper) which basically wraps a pointer and mimics a reference. Therefore it can be used in View.

This would also be a motivation to change the unique_ptr of Card to a std::refrence_wrapper and let the State own the Card. Here we can also argue the same, that a View should not own the Card but the state.

include/controller/view/view_item.hpp
```diff
@@ -8,5 +8,6 @@ namespace controller {
 // Forward declarations to avoid circular dependencies
 struct Card; // Card used in ViewItem, but Card itself includes ViewItem, so we need a forward declaration here
 
-using ViewItem = std::variant<std::unique_ptr<Card>, Button, Text>;
+// Button shared with MenuState buttons_ list
+using ViewItem = std::variant<std::unique_ptr<Card>, std::shared_ptr<Button>, Text>;
```
5tefan543: Refactor `std::shared_ptr<Button>` to `std::reference_wrapper<Button>` and if possible also `std::unique_ptr<Card>` to `std::reference_wrapper<Card>`.

include/controller/state/state.hpp
```diff
@@ -17,22 +17,30 @@ struct BaseState {
...
 enum class MenuType { MainMenu, PauseMenu, GameOverMenu };
 
 struct MenuState : public BaseState {
     // Add menu-specific state variables here
-    MenuType type;
-    std::size_t selectedButtonIndex = 0;
+    const MenuType type;
 
-    static std::unique_ptr<MenuState> createMenu(MenuType menuType);
+    static std::unique_ptr<MenuState> createMenu(const MenuType menuType);
```
5tefan543, line 43: remove `const` for the parameter since it does not add anything useful since the menuType is copied .

src/controller/state/state.cpp
```diff
@@ -3,64 +3,80 @@
...
+MenuState::MenuState(MenuType type) : type(type)
+{
+    initView();
+}
+
 std::unique_ptr<MenuState> MenuState::createMenu(MenuType menuType)
 {
-    auto menu = std::make_unique<MenuState>();
-    menu->type = menuType;
+    auto menu = std::unique_ptr<MenuState>(new MenuState(menuType));
```
5tefan543: Refactor to simply return

tests/controller/state/state_test.cpp
```diff
@@ -4,6 +4,36 @@
 
 using namespace controller;
 
+enum INPUT { UP, DOWN, CONFIRM, NONE, LEFT, RIGHT };
+StateTransitionAction applyInput(std::unique_ptr<MenuState> &state, const INPUT in)
```
5tefan543: Could you move this in `shared/util.hpp` since that could be useful for other tests.

include/controller/state/state.hpp
5tefan543: change to `class` and also all child classes

tests/controller/state/state_test.cpp
5tefan543: Remove additional assert here, since only Menu creation should be verified.

tests/controller/state/state_test.cpp
```diff
@@ -62,171 +89,91 @@ TEST_CASE("ProgressionStoreState::createStore constructs store state with expect
...
+        {
+            INFO("confirm pressed with initial button selection triggers ReplaceCurrentWithGameplay");
+            REQUIRE(applyInput(state, CONFIRM) == StateTransitionAction::ReplaceCurrentWithGameplay);
+        }
+        {
+            INFO("down input returns no state transition action");
+            REQUIRE(applyInput(state, DOWN) == StateTransitionAction::None);
+        }
+        {
+            INFO("confirm exits game when quit button is selected (through previous down input)");
```
5tefan543: I don't think it is good to have a test section that depends on the above execution here. If someone moves the sections around then the tests break.

Better would be to always start fresh and perform multiple actions in one section like this


```
{
  INFO("confirm exits game when quit button is selected");
  REQUIRE(applyInput(state, DOWN) == StateTransitionAction::None);
  REQUIRE(applyInput(state, CONFIRM) == StateTransitionAction::ReplaceAllStatesWithExit);
}
```

src/controller/state/state.cpp
```diff
@@ -193,11 +211,6 @@ StateTransitionAction GameplayState::update(const InputState &input, DebugContex
     return controller::StateTransitionAction::None;
 }
```
5tefan543, line 239: Also add an override like 

```
const View &GameplayState::getView()
{
    return game.getView(view_);
}
```

since this should be part of this refactoring.

I understand that you maybe don't want to touch the Game code since @RZacchia does something there. But this refactor is so essential and a very small change to Game that it is not valid in my opinion to simply ignore the game.

If we now merge this without that part in main and @RZacchia updates his branch then his game rendering won't work since Game::getView is not called anymore. Then he has to touch your refactored logic in the controller to make it work again.

5tefan543 (changes requested):

include/controller/state/state.hpp
51th08: I agree. Will look into it.

include/controller/state/state.hpp
51th08: I'll change it if you want but out of curiosity:

My thoughs about const:
Why not make everything const that can be?
It makes the intention of the method interface more clear.
It captures the intend of using the info without modification. Could even prevent bugs and calls for attention (with const modif. errors) when changes are suddenly made (e.g. refactoring method logic - maybe interface needs changing as well?).
In this instance it might be on the verbose site but it goes with the convention/intent of the previous points.

tests/controller/state/state_test.cpp
51th08: Since all these actions depend on each other this would be the logical conclusion:
```cpp
        {
            INFO("confirm pressed with initial button selection triggers ReplaceCurrentWithGameplay");
            REQUIRE(applyInput(state, CONFIRM) == StateTransitionAction::ReplaceCurrentWithGameplay);
            INFO("down input returns no state transition action");
            REQUIRE(applyInput(state, DOWN) == StateTransitionAction::None);
            INFO("confirm exits game when quit button is selected (through previous down input)");
            REQUIRE(applyInput(state, CONFIRM) == StateTransitionAction::ReplaceAllStatesWithExit);
            INFO("no relevant input returns None");
            REQUIRE(applyInput(state, NONE) == StateTransitionAction::None);
            INFO("up input returns no state transition action");
            REQUIRE(applyInput(state, UP) == StateTransitionAction::None);
            INFO("confirmPressed triggers ReplaceCurrentWithGameplay (through previous up input)");
            REQUIRE(applyInput(state, CONFIRM) == StateTransitionAction::ReplaceCurrentWithGameplay);
        }
```
I understood Catch2's doc like that this way all INFO()s apply to all REQUIRE()s because they are in the same scope. Am I wrong? We could simply delete all INFO()s or write one INFO("Test all internal menu states with corresponding inputs") but that throws away error msg details.

There's also the AND_THEN() option but it seemed to require nesting each new AND_THEN() inside the
previous AND_THEN() scope which would make this much more unwieldy.

src/controller/state/state.cpp
51th08: Do I understand you correctly that game should own its view and GameplayState has only a reference to it?

If so, I think that goes against the whole (Base)State architecture we build up to now.
What do you think of [38-game-view-handling-suggestion](https://github.com/5tefan543/uibk_cpp_project/tree/38-game-view-handling-suggestion)?

include/controller/state/state.hpp
5tefan543: After researching this a bit more, I generally agree with you. But you also need to make the parameter in the implementation `const` (which it is currently not). The `const` in the header alone does not do anything.

src/controller/state/state.cpp
```diff
@@ -3,64 +3,79 @@
...
+{
+    return view_;
+}
+
+MenuState::MenuState(MenuType type) : type(type)
+{
+    initView();
+}
+
 std::unique_ptr<MenuState> MenuState::createMenu(MenuType menuType)
```
5tefan543: make parameter `const`

tests/controller/state/state_test.cpp
5tefan543: I think my concern is less about Catch2 itself and more about the general test structure. Right now the current version tests the whole MainMenu flow in one combined scenario, and the cases are order-dependent because they all mutate the same state. The bare scopes look like separate checks, but they’re really just one sequential scenario.

I would prefer testing each menu behavior separately, with a fresh state per case, so each check stands on its own and isn’t coupled to previous inputs. Something like this:

```
TEST_CASE("Main menu update returns correct actions")
{
    auto state = MenuState::createMenu(MenuType::MainMenu);

    SECTION("confirm on initial selection starts gameplay")
    {
        REQUIRE(applyInput(state, CONFIRM) == StateTransitionAction::ReplaceCurrentWithGameplay);
    }

    SECTION("down changes selection so confirm exits")
    {
        REQUIRE(applyInput(state, DOWN) == StateTransitionAction::None);
        REQUIRE(applyInput(state, CONFIRM) == StateTransitionAction::ReplaceAllStatesWithExit);
    }

    SECTION("up from quit selection returns to start game")
    {
        REQUIRE(applyInput(state, DOWN) == StateTransitionAction::None);
        REQUIRE(applyInput(state, UP) == StateTransitionAction::None);
        REQUIRE(applyInput(state, CONFIRM) == StateTransitionAction::ReplaceCurrentWithGameplay);
    }

    SECTION("irrelevant input returns None")
    {
        REQUIRE(applyInput(state, NONE) == StateTransitionAction::None);
    }
}
```

In my opinion this is also easier to read.

If we intentionally want to test a sequence, I’d group that into one explicit section. Otherwise, for distinct behaviors, using fresh state per section makes the tests more robust and easier to refactor.

src/controller/state/state.cpp
5tefan543: I think there is a misunderstanding about what I meant.

I did not mean that Game should own the view, or that GameplayState should only hold a reference to it.

My intention was that GameplayState still owns `view_` through BaseState, just like before. The missing part of the refactor in my opinion is that GameplayState should override `getView()` and delegate to `Game::getView(...)`, so gameplay rendering still goes through Game.

So the idea was not to store a reference to the state-owned view inside Game, but only to pass the view to `Game::getView(...)` when the controller asks for the current gameplay view.

That way, view ownership stays in the state layer, Game does not keep a long-lived reference to render data, and the gameplay path is still updated as part of this refactor.

In this context it would maybe also make sense to refactor `Game::getView` to `Game::updateView`

5tefan543 (changes requested):

include/controller/state/state.hpp
51th08: Ah, missed that const

src/controller/state/state.cpp
51th08: Ah yeah that sounds good :+1:

5tefan543 (approved):

PR #46, opened 2026-05-01
20 integrate json library glaze implement save logic

### LINKED ISSUE #20: Integrate JSON Library Glaze & Implement save logic
- Decide save format -> JSON
- Implement minimal save logic

### LINKED ISSUE #48: Camera not correctly following player anymore
I noticed that the camera now does not follow the player anymore if it goes to the left or down.

_Originally posted by @5tefan543 in https://github.com/5tefan543/uibk_cpp_project/pull/46#discussion_r3177841741_

RZacchia:


include/controller/persistence/config_game.hpp
5tefan543, line 1: Can we remove all the initializations here like

`int width = 1920;`

or

`std::string playerTexturePath = "assets/characters/player_1_";`

since not having them is the entire reason of a config right?

include/controller/persistence/config_game.hpp
```diff
@@ -0,0 +1,27 @@
...
+    std::string mapTexturePath = "assets/map/map_1.png";
+    std::string fontPath = "assets/font/BigBlueTerm_Nerd_Font/BigBlueTerm437NerdFont-Regular.ttf";
+};
+
+struct GameConfig {
+    int initialStage = 1;
+    int initialWave = 1;
+    int initialCurrency = 0;
+    WindowConfig windowConfig;
+    AssetConfig assetConfig;
```
5tefan543, line 24: If you nest the `WindowConfig` and the `AssetConfig` within the `GameConfig`, shouldn’t this structure be reflected in the JSON file as well?

config/saved-game.json
5tefan543, line 1: Should we put this file in .gitignore?

Why is it called saved-game.json? This should store the `PersistedGame` right?

include/controller/persistence/leaderboard.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#include <string>
+
+struct LeaderboardEntry {
+    std::string playerName;
+    int score;
+    int wave;
```
5tefan543: Do we need `stage` here as well?

include/controller/persistence/persistence_manager.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+#include "controller/persistence/config_game.hpp"
+#include "controller/persistence/persisted_game.hpp"
+#include "controller/state/state_manager.hpp"
+#include "game/ecs/registry.hpp"
+namespace controller {
```
5tefan543, line 8: refactor to only needed includes

```
#include "controller/persistence/config_game.hpp"
#include "controller/persistence/persisted_game.hpp"
#include <vector>
#include <string>
```

include/controller/persistence/persistence_manager.hpp
```diff
@@ -0,0 +1,25 @@
...
+    PersistenceManager();
+    ~PersistenceManager();
+
+    void saveGame(const controller::PersistedGame &persistedGame);
+    void loadGame(controller::PersistedGame &persistedGame);
+    bool hasSavedGame();
+    void deleteSave();
+
+    void storeLeaderboardEntry(const std::string &playerName, int score);
+    std::vector<std::pair<std::string, int>> getLeaderboardEntries(int topN);
```
5tefan543: I am confused. Should this not return objects of type `LeaderboardEntry`.

What does `topN` mean? -> Better naming!

include/controller/state/state.hpp
```diff
@@ -47,10 +47,14 @@ class MenuState : public BaseState {
 };
 
 class GameplayState : public BaseState {
+    bool loadedFromSave_ = false;
+
   public:
     game::Game game;
 
     static std::unique_ptr<GameplayState> createGameplay();
+    static std::unique_ptr<GameplayState> createLoadedGameplay();
```
5tefan543, line 55: maybe refactor to 

`static std::unique_ptr<GameplayState> createGameplay(bool loadPersistedGame);`

include/controller/state/state.hpp
```diff
@@ -47,10 +47,14 @@ class MenuState : public BaseState {
...
+    bool didLoadFromSave() const;
```
5tefan543: Unexpected naming -> `isLoadedFromPersistedGame()`

include/game/game.hpp
```diff
@@ -22,10 +23,15 @@ class Game {
     InputSystem inputSystem_;
     MovementSystem movementSystem_;
 
+    int stage_ = 1;
+    int wave_ = 1;
```
5tefan543, line 29: Ensure this is consistent with `GameDebugSettings.stage` and `GameDebugSettings.wave` for debugging.

For example synchronize `DebugContext.GameDebugSettings` once when loading game and then within game the local variables should be modifiable via debug ui.

include/game/game.hpp
```diff
@@ -37,6 +43,8 @@ class Game {
     ~Game();
 
     GameDebugSession &getDebugSession();
+    void loadFromPersistedGame(const controller::PersistedGame &persistedGame);
```
5tefan543, line 48: Does `PersistedGame` really belong to namespace `controller`. I would put it in `game`

src/controller/state/state.cpp
```diff
@@ -49,6 +53,13 @@ StateTransitionAction MenuState::update(const InputState &input, [[maybe_unused]
                 stateTransitionAction = StateTransitionAction::ReplaceCurrentWithGameplay;
                 break;
             case 1:
+                if (buttons_.size() == 3) {
```
5tefan543, line 56: We currently cannot save the game, right? So there is currently no way to see this button or did I miss something. I mean this is not yet implemented, right?

If so, then it would be great to add an option for this save logic in the DebugUI with a button to save (also when switching to a different stage or wave in the DebugUI) to see if the save logic works as intended.

src/controller/state/state.cpp
```diff
@@ -123,6 +134,9 @@ void MenuState::initView()
 {
     switch (type) {
     case MenuType::MainMenu: {
+        PersistenceManager persistenceManager;
```
5tefan543: Should we always recreate the `PersistenceManager` every frame?

src/controller/state/state.cpp
```diff
@@ -136,13 +150,23 @@ void MenuState::initView()
...
         Button &quitButton = buttons_.emplace_back(Button());
         quitButton.text.text = std::string("Quit");
         quitButton.width = 300.0f;
-        quitButton.centerOffsetY = 100;
+        quitButton.centerOffsetY = hasSavedGame ? 200.0f : 100.0f;
 
         mainMenuCard.items.push_back(title);
         mainMenuCard.items.push_back(startGameButton);
+        if (hasSavedGame) {
+            mainMenuCard.items.push_back(buttons_[1]);
```
5tefan543: Refactor to not access with indexing.

src/controller/controller.cpp
```diff
@@ -1,11 +1,13 @@
 #include "controller/controller.hpp"
+#include "controller/persistence/persistence_manager.hpp"
 #include <iostream>
 
 namespace controller {
 
 Controller::Controller()
 {
     std::cout << "Controller constructed" << std::endl;
+    PersistenceManager persistenceManager;
```
5tefan543: Is this one needed?

src/game/systems/camera_system.cpp
```diff
@@ -33,6 +33,9 @@ void CameraSystem::update(Registry &registry)
     // Clamp camera to map boundaries
     camera.x = std::max(-64.0f, std::min(camera.x, map.width));
     camera.y = std::max(-64.0f, std::min(camera.y, map.height));
+
+    camera.x = std::min(camera.x, 160.0f);
+    camera.y = std::min(camera.y, 250.0f);
```
5tefan543: I noticed that the camera now does not follow the player anymore if it goes to the left or down.

src/controller/persistence/persistence_manager.cpp
```diff
@@ -0,0 +1,192 @@
...
+namespace fs = std::filesystem;
+
+namespace {
+
+const fs::path configDir{"config"};
+const fs::path saveFilePath = configDir / "saved-game.json";
+const fs::path leaderboardFilePath = configDir / "leaderboard.json";
+const fs::path configFilePath = configDir / "game-config.json";
+
+struct LeaderboardEntry {
```
5tefan543: This already exists doesn't it?

src/controller/persistence/persistence_manager.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include <algorithm>
+#include <filesystem>
+#include <fstream>
+#include <glaze/glaze.hpp>
+#include <iostream>
+#include <sstream>
+
+namespace fs = std::filesystem;
+
+namespace {
```
5tefan543: What is the reason for not using `controller` namspace?

Maybe move these glaze helper methods to a static helper struct?

src/controller/persistence/persistence_manager.cpp
```diff
@@ -0,0 +1,192 @@
...
+struct LeaderboardEntry {
+    std::string playerName;
+    int score = 0;
+};
+
+template <typename T>
+bool writeJsonToFile(const T &value, const fs::path &path)
+{
+    std::string json;
+    if (const auto err = glz::write_json(value, json)) {
```
5tefan543: `err` is ignored? Should we print the error?

5tefan543: Could we also use `glz::write_file_json` ?

src/controller/persistence/persistence_manager.cpp
```diff
@@ -0,0 +1,192 @@
...
+{
+    std::ifstream in(path);
+    if (!in) {
+        return false;
+    }
+
+    std::stringstream buffer;
+    buffer << in.rdbuf();
+
+    if (const auto err = glz::read_json(value, buffer.str())) {
```
5tefan543: Also here `err` is ignored

Can we use `glz::read_file_json` ?

src/controller/persistence/persistence_manager.cpp
```diff
@@ -0,0 +1,192 @@
...
+        std::cerr << "Failed to deserialize JSON for " << path << std::endl;
+        return false;
+    }
+
+    return true;
+}
+
+std::vector<LeaderboardEntry> readLeaderboardEntriesFromDisk()
```
5tefan543: Does this then belong to the glaze helper methods? This is not within the `controller` namespace.

5tefan543 (changes requested):

include/controller/persistence/config_game.hpp
RZacchia: i leave them as fallback, else they init willl not work if the config is misformed or missing

include/controller/persistence/leaderboard.hpp
RZacchia: no

src/controller/state/state.cpp
RZacchia: if u add an empty save file you can see the button

include/controller/persistence/leaderboard.hpp
5tefan543: Why not?

src/controller/state/state.cpp
5tefan543: What empty file?

`saved-game.json` ?

include/controller/persistence/config_game.hpp
5tefan543: But the config should not be missing. If the config is missing, the app should not start.

src/controller/state/state.cpp
5tefan543: Ah now I know why the button was not visible for me. I started the app within `build/debug` and there is no `config` present.

So we need to copy the `config` to the build directory. In the same way as we already do for the `assets` folder:

```
add_custom_command(TARGET roguelike_app POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy_directory
            ${CMAKE_SOURCE_DIR}/assets
            $<TARGET_FILE_DIR:roguelike_app>/assets
)
```

src/controller/persistence/persistence_manager.cpp
RZacchia: error is printed the line below

src/game/systems/camera_system.cpp
RZacchia: <insert it works on my machine joke> yeah i realized that just now when trying it on the laptop

src/controller/persistence/persistence_manager.cpp
RZacchia: if it is in controller namespace we get issues. yeah i will move it into another namespace

src/controller/persistence/persistence_manager.cpp
RZacchia: yes we could, but I would stay that way for now, should something change in the way we store/read (e.g. checksums etc)

src/controller/persistence/persistence_manager.cpp
RZacchia: same as above

src/controller/state/state.cpp
RZacchia: we do not make it every frame, check the debug output it only gets constructed once when loading a game and saving the game and destroyed afterwards

src/controller/persistence/persistence_manager.cpp
5tefan543: What issues?

src/controller/persistence/persistence_manager.cpp
5tefan543: No the `err` variable is ignored and not used in the next line when printing the error. Also error handling can be improved.

I meant something like that:

```
template <typename T>
bool writeJsonToFile(const T& value, const fs::path& path)
{
    std::string json;

    if (const auto err = glz::write_json(value, json)) {
        std::cerr << "Failed to serialize JSON for " << path
                  << ": " << glz::format_error(err, json) << '\n';
        return false;
    }

    std::error_code ec;
    fs::create_directories(path.parent_path(), ec);
    if (ec) {
        std::cerr << "Failed to create directory for " << path
                  << ": " << ec.message() << '\n';
        return false;
    }

    std::ofstream out(path);
    if (!out) {
        std::cerr << "Failed to open file for writing: " << path << '\n';
        return false;
    }

    out << json;

    if (!out) {
        std::cerr << "Failed to write JSON to file: " << path << '\n';
        return false;
    }

    return true;
}
```

Similar for the reading part.

src/controller/persistence/persistence_manager.cpp
5tefan543: Refactor also error handling, something like:

```
template <typename T>
bool readJsonFromFile(T& value, const fs::path& path)
{
    std::ifstream in(path);
    if (!in) {
        std::cerr << "Failed to open JSON file for reading: " << path << '\n';
        return false;
    }

    std::string json{
        std::istreambuf_iterator<char>{in},
        std::istreambuf_iterator<char>{}
    };

    if (!in.eof() && in.fail()) {
        std::cerr << "Failed to read JSON file: " << path << '\n';
        return false;
    }

    if (const auto err = glz::read_json(value, json)) {
        std::cerr << "Failed to deserialize JSON for " << path
                  << ": " << glz::format_error(err, json) << '\n';
        return false;
    }

    return true;
}
```

src/controller/persistence/persistence_manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+    fs::remove(saveFilePath, ec);
+    if (ec) {
+        std::cerr << "Failed to delete save file: " << saveFilePath << std::endl;
+    }
+}
+
+void PersistenceManager::storeLeaderboardEntry(const std::string &playerName, int score)
+{
+    auto entries = readLeaderboardEntriesFromDisk();
+    entries.push_back({playerName, score});
```
5tefan543: LeaderboardEntry::wave is not initialized here (see warnings)

src/controller/state/state.cpp
5tefan543: Ah right, I still had it in my head that the view gets reconstructed every frame, sorry :D

5tefan543: But in general, wouldn’t it still be cleaner to add the `PersistenceManager` as a member? Since we’ll need it again in `update() `for saving the game anyway, right?

Or create the `PersistenceManager` once in the controller (since we need to load the `GameConfig` somewhere anyway) and pass it where needed. This would generally be better if the `PersistenceManager`  also caches the game in the future?

tests/controller/state/state_manager_test.cpp
```diff
@@ -1,8 +1,39 @@
...
+
+fs::path createTempTestDirectory()
+{
+    const auto uniquePart = std::chrono::steady_clock::now().time_since_epoch().count();
+    const fs::path dir = fs::temp_directory_path() / ("roguelike-state-manager-test-" + std::to_string(uniquePart));
+    fs::create_directories(dir);
+    return dir;
+}
+
+} // namespace
```
5tefan543: Move to own file (e.g. `tests/shared/util.cpp`)

tests/controller/state/state_test.cpp
```diff
@@ -1,9 +1,51 @@
...
+    fs::path oldPath_;
+};
+
+fs::path createTempTestDirectory()
+{
+    const auto uniquePart = std::chrono::steady_clock::now().time_since_epoch().count();
+    const fs::path dir = fs::temp_directory_path() / ("roguelike-state-test-" + std::to_string(uniquePart));
+    fs::create_directories(dir);
+    return dir;
+}
```
5tefan543: Duplicate code? Already defined in `state_manager_test.cpp`. Even more reason to move to shared test util.

tests/controller/state/state_manager_test.cpp
```diff
@@ -154,6 +185,38 @@ TEST_CASE("applyAction ReplaceCurrentWithGameplay replaces current state with ga
...
+    REQUIRE(gameplayState->isLoadedFromPersistedGame());
+
+    const PersistedGame loaded = gameplayState->game.getPersistedGame();
+    REQUIRE(loaded.stage == 9);
+    REQUIRE(loaded.wave == 4);
+    REQUIRE(loaded.currency == 777);
+    REQUIRE(loaded.playerStats.speed == 360.0f);
+
+    std::error_code ec;
+    fs::remove_all(testDir, ec);
```
5tefan543: Should the cleanup not happen in the destructor of `ScopedCurrentPath` ?

The problem with that is that we will forget to cleanup in the future. And I also noticed that if a test fails, the remaining code is no longer executed, meaning the test directories are not cleaned up on failure. (does not happen with RAII)

Also why split the creation of the test folder in a two way process?
1. First the testDir folder is created?
2. Then the current path is changed to the testDir?

Why not simply handle everything in the class object? For example like this?

```
class ScopedTestDirectory {
  public:
    ScopedTestDirectory()
        : path_(createUniqueTempDirectory()),
          oldPath_(fs::current_path())
    {
        fs::current_path(path_);
    }

    ~ScopedTestDirectory()
    {
        fs::current_path(oldPath_);

        std::error_code ec;
        fs::remove_all(path_, ec);
    }

    [[nodiscard]]
    const fs::path &path() const
    {
        return path_;
    }

  private:
    static fs::path createUniqueTempDirectory()
    {
        const auto uniquePart =
            std::chrono::steady_clock::now().time_since_epoch().count();

        const fs::path dir =
            fs::temp_directory_path() /
            ("roguelike-test" + std::to_string(uniquePart));

        fs::create_directories(dir);
        return dir;
    }

    fs::path path_;
    fs::path oldPath_;
};
```

5tefan543 (changes requested):

src/game/systems/camera_system.cpp
5tefan543: Created Bug #48 (do not fix in this PR)

include/game/game.hpp
5tefan543: Added new issue #49 (not handle in this PR)

5tefan543 (changes requested):
What is still missing, and should be part of this PR, is the loading and handling of the `GameConfig`.

The current implementation neither loads the config nor uses it anywhere in the codebase (e.g. the game does not read `playerTexturePath` from the config).

The `GameConfig` should be loaded once in the controller and then passed to the parts of the application that require access to it, similar to how `InputState` and `DebugContext` are handled.

src/controller/controller.cpp
```diff
@@ -7,6 +8,9 @@ Controller::Controller()
 {
     std::cout << "Controller constructed" << std::endl;
     stateManager_.push(MenuState::createMenu(MenuType::MainMenu));
+
+    PersistenceManager persistenceManager;
+    gameConfig_ = persistenceManager.loadConfig();
```
5tefan543: Maybe the `GameConfig` should load itself in the ctor?

What do you think about making `PersistenceManager` static?

include/controller/persistence/leaderboard.hpp
RZacchia: the stage has nothing to do with the progress of the game and is not needed in the leaderboard stats

e.g. 
stage 1 - wave 1
stage 1 - wave 2
....
progression store
stage 2 - wave 6
stage 2 - wave 7
...
progression store
stage 1 - wave 11
stage 1 - wave 12
...
...

5tefan543: Ah now that makes sense. I thought a wave starts again at 1 in a new stage. ;D

include/controller/persistence/persisted_game.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+namespace controller {
+
+// only temporary struct
+// should be replaced by actual components and systems that manage them
```
5tefan543: Is this comment outdated?

Because the struct is not temporary, is it?

include/controller/persistence/serializer.hpp
```diff
@@ -0,0 +1,122 @@
...
+    static inline const std::filesystem::path configDir{"config"};
+    static inline const std::filesystem::path saveFilePath = configDir / "persisted-game.json";
+    static inline const std::filesystem::path leaderboardFilePath = configDir / "leaderboard.json";
+    static inline const std::filesystem::path configFilePath = configDir / "game-config.json";
+
+    template <typename T>
+    static bool writeJsonToFile(const T &value, const std::filesystem::path &path)
+    {
+        std::string json;
+        if (const auto err = glz::write_json(value, json)) {
```
5tefan543, line 75: The `err` variable is ignored and not used in the next line when printing the error. Also error handling can be improved.

For example like that

```
template <typename T>
static bool writeJsonToFile(const T &value, const std::filesystem::path &path)
{
    std::string json;
    if (const auto err = glz::write_json(value, json)) {
        std::cerr << "Failed to serialize JSON for " << path
                  << ": " << glz::format_error(err, json) << '\n';
        return false;
    }

    const auto parentPath = path.parent_path();
    if (!parentPath.empty()) {
        std::error_code ec;
        std::filesystem::create_directories(parentPath, ec);

        if (ec) {
            std::cerr << "Failed to create directories for " << path
                      << ": " << ec.message() << '\n';
            return false;
        }
    }

    std::ofstream out(path);
    if (!out) {
        std::cerr << "Failed to open file for writing: " << path << '\n';
        return false;
    }

    out << json;

    if (!out) {
        std::cerr << "Failed to write JSON to file: " << path << '\n';
        return false;
    }

    return true;
}
```

Similar for the reading part.

include/controller/persistence/serializer.hpp
```diff
@@ -0,0 +1,122 @@
...
+        std::stringstream buffer;
+        buffer << in.rdbuf();
+
+        if (const auto err = glz::read_json(value, buffer.str())) {
+            std::cerr << "Failed to deserialize JSON for " << path << std::endl;
+            return false;
+        }
+
+        return true;
+    }
```
5tefan543, line 123: Also this can be improved, e.g.:

```
template <typename T>
static bool readJsonFromFile(T &value, const std::filesystem::path &path)
{
    std::ifstream in(path);
    if (!in) {
        return false;
    }

    std::string json{
        std::istreambuf_iterator<char>{in},
        std::istreambuf_iterator<char>{}
    };

    if (!in.good() && !in.eof()) {
        std::cerr << "Failed to read JSON file: " << path << '\n';
        return false;
    }

    if (const auto err = glz::read_json(value, json)) {
        std::cerr << "Failed to deserialize JSON for " << path
                  << ": " << glz::format_error(err, json) << '\n';
        return false;
    }

    return true;
}
```

include/controller/persistence/serializer.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    static std::vector<LeaderboardEntry> readLeaderboardEntriesFromDisk()
```
5tefan543: Does this logically not belong more to the `PersistenceManager` ?

include/controller/controller.hpp
```diff
@@ -18,6 +19,7 @@ class Controller {
     void update(const InputState &input, float dt);
     BaseState &getCurrentState();
     DebugContext &getDebugContext();
+    GameConfig &getGameConfig();
```
5tefan543: Not needed anymore. Also the include ;)

tests/controller/state/state_test.cpp
```diff
@@ -115,6 +133,43 @@ TEST_CASE("Main menu mouse input returns correct actions")
...
+
+    //         const view::Button &startButton = ViewItemAccessor::as<const view::Button>(card.items[1]);
+    //         const view::Button &loadButton = ViewItemAccessor::as<const view::Button>(card.items[2]);
+    //         const view::Button &quitButton = ViewItemAccessor::as<const view::Button>(card.items[3]);
+
+    //         REQUIRE(startButton.text.text == "Start Game");
+    //         REQUIRE(loadButton.text.text == "Load Game");
+    //         REQUIRE(quitButton.text.text == "Quit");
+    //     }
+}
```
5tefan543: What about this?

tests/controller/state/state_test.cpp
```diff
@@ -405,6 +461,32 @@ TEST_CASE("MenuState::getView returns expected view")
...
+    //         const view::Button &startButton = ViewItemAccessor::as<const view::Button>(card.items[1]);
+    //         REQUIRE(startButton.text.text == "Start Game");
+
+    //         const view::Button &loadButton = ViewItemAccessor::as<const view::Button>(card.items[2]);
+    //         REQUIRE(loadButton.text.text == "Load Game");
+
+    //         const view::Button &quitButton = ViewItemAccessor::as<const view::Button>(card.items[3]);
+    //         REQUIRE(quitButton.text.text == "Quit");
+    //     }
+
```
5tefan543: What about this?

tests/controller/persistence/persistence_manager_test.cpp
5tefan543, line 1: Why is this in `tests/utils/persistence/` ?

Move to `tests/controller/persistence/`

5tefan543 (changes requested):

5tefan543 (changes requested):
Open Comments

include/controller/state/state.hpp
RZacchia: no, refactored createGameplay to createNewGameplay, to keep method names descriptive

include/game/game.hpp
RZacchia: imo PersistedGame is a dataclass of PersistenceManager, so i keep it in this namespace

src/controller/persistence/persistence_manager.cpp
RZacchia: kept old glz method, since some errors get ommited if we use read_file_json

RZacchia: kept old glz method, since some errors get ommited if we use write_file_json

5tefan543 (approved):

PR #51, opened 2026-05-08
DebugUI: improve entity selection

### LINKED ISSUE #50: DebugUI: Selected Entity visible
Selecting an entity in the DebugUI should be visible on screen. The selected entity should somehow be marked as selected.

5tefan543:


51th08:
How about
```cpp
void DebugUI::renderEcsManagement(game::GameDebugSession &gameSession)
{
    if (ImGui::CollapsingHeader("ECS", ImGuiTreeNodeFlags_DefaultOpen)) {

        ImGui::Text("Entity count: %zu", gameSession.registry.entities().size());
        ImGui::Text("Select on screen: Ctrl + left-mouse-btn");
```
such that I don't forget how to select entities directly on screen

src/game/systems/debug_selection_system.cpp
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    for (auto entity : registry.view<Map>()) {
+        const Map &map = registry.getComponent<Map>(entity);
+
+        const float x = map.x * map.scale;
+        const float y = map.y * map.scale;
+        const float width = map.width * map.scale;
+        const float height = map.height * map.scale;
```
51th08, line 77: Why to we need a scale for map & sprites? Couldn't it be applied directly to to their respective width height?
- It would reduce code complexity.
- One would not have to remember applying scaling factors when using e.g. a sprites width/height.
- It feels a bit strange to start scaling stuff since we're transitioning to a fixed internal resolution/grid.

5tefan543: I didn't introduced the scaling. This was already there.

Maybe we don't need that / should not use that anymore with the new fixed grid design.

@RZacchia What do you think about that?

RZacchia: I am unsure about it in generel, but leave it for now. This should be handled by a different issue.
If you just remove it now, the map will shrink in comparission to the character.

I drew the map 1920 x 1080, but we wanted that the map is larger, so that we do not see the whole map all the time in fullscreen => scaling

RZacchia (approved):

PR #52, opened 2026-05-08
Fix coord system and render view

51th08:


src/ui/ui.cpp
```diff
@@ -43,7 +58,7 @@ bool UI::isOpen() const
 controller::InputState UI::pollInput()
 {
     inputState_ = inputHandler_.pollInput(window_);
-    return inputState_;
+    return inputState_; // return implicit copy of internal inputState_
```
5tefan543: Should we return a reference here?

51th08: Ah, yes we could. Probably even a "const reference!" ;)

In [Moving to internal grid representation and more stuff](https://github.com/5tefan543/uibk_cpp_project/pull/52/commits/ef76f7e6c816cdc75ab7bee4d4cb0c6dfcd0232a) I accidentally made a one-liner out of that without realizing we need the copy/reference. This was just undoing that.

src/ui/ui.cpp
```diff
@@ -43,16 +83,18 @@ bool UI::isOpen() const
...
 }
 
-void UI::render(const controller::View &view, controller::DebugContext &debug)
+void UI::render(const view::View &view, controller::DebugContext &debug)
 {
+    if (inputState_.windowResized) {
+        setSfmlView();
+    }
     // 1. Start ImGui frame
```
5tefan543: Maybe refactor comments?

// 1. Reset sfml view on resize
// 2. Start ImGui frame
... (with updated numbering)

51th08: Comments are added/adjusted [e68516a](https://github.com/5tefan543/uibk_cpp_project/pull/52/commits/e68516a4b95b5a62f15ed063b12b2a56fa50e525)

src/ui/ui.cpp
51th08: Done [f1d5efc](https://github.com/5tefan543/uibk_cpp_project/pull/52/commits/f1d5efcf616f47fa11efb5761304da13a035ca6c)

5tefan543 (approved):

PR #58, opened 2026-05-14
47 implement gameplay wave transitions

### LINKED ISSUE #47: Implement gameplay wave transitions
- Configurable timer for waves
- AutoSave on Wave change
- Configurable number of waves before transistion to progress store
- Going into next stage after store
- auto delete on game over
- able to load into correct wave with save file

TLDR: so that the game can already move through its cycles (parameters via gameConfig) without the cycles themselves

RZacchia:
closes #47


include/controller/state/state.hpp
```diff
@@ -60,7 +61,17 @@ class GameplayState : public BaseState {
...
+    std::deque<view::Button> buttons_;
+    std::deque<view::Card> cards_;
+    std::deque<view::Text> texts_;
+    std::size_t selectedButtonId_ = 0;
+
+    void initView();
+    std::optional<std::size_t> getHoveredButtonId(const InputState &input) const;
+
   public:
+    ProgressionStoreState();
```
5tefan543: make private same as in MenuState

src/controller/state/state.cpp
```diff
@@ -317,12 +320,98 @@ std::unique_ptr<ProgressionStoreState> ProgressionStoreState::createStore()
     return std::make_unique<ProgressionStoreState>();
 }
 
+std::optional<std::size_t> ProgressionStoreState::getHoveredButtonId(const InputState &input) const
```
5tefan543: Duplicate code!

Should we move this logic to a static MouseUtil ? For example like:

```
static std::optional<std::size_t> MouseUtil::getHoveredButtonId(const InputState &input, const std::deque<view::Button> &buttons)
{
    for (std::size_t idx = 0; idx < buttons.size(); idx++) {
        const view::Button &button = buttons[idx];
        const bool insideX = input.mouseGridX >= button.gridX && input.mouseGridX <= (button.gridX + button.width);
        const bool insideY = input.mouseGridY >= button.gridY && input.mouseGridY <= (button.gridY + button.height);

        if (insideX && insideY) {
            return idx;
        }
    }

    return std::nullopt;
}
```

include/game/game.hpp
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
+#include "controller/debug/debug_context.hpp"
```
5tefan543: Not needed

include/game/game.hpp
```diff
@@ -11,21 +13,28 @@
 #include "game/ecs/systems/movement_system.hpp"
 #include "view/view.hpp"
 
+#include <chrono>
```
5tefan543: Not needed

include/game/game.hpp
```diff
@@ -11,21 +13,28 @@
...
+
 namespace game {
 
 class Game {
   private:
     Registry registry_;
     GameDebugSession debugSession_{registry_};
+    controller::GameConfig config_;
```
5tefan543, line 21: Why do we store a copy of the config here?

If we later want to edit the config via the DebugUI, that is more work. GameConfig should never be stored and always directly loaded with `controller::PersistenceManager::getConfig();` when actually needed.

include/game/game.hpp
```diff
@@ -11,21 +13,28 @@
...
+    float currentWaveDuration_;
 
     AnimationSystem animationSystem_;
     CameraSystem cameraSystem_;
     InputSystem inputSystem_;
     MovementSystem movementSystem_;
     DebugSelectionSystem debugSelectionSystem_;
 
+    int waveDurationSeconds_;
+    int wavesPerStage_;
```
5tefan543: always read from GameConfig

src/game/game.cpp
```diff
@@ -18,10 +19,12 @@ namespace game {
 Game::Game()
 {
     std::cout << "Game constructed" << std::endl;
+    config_ = controller::PersistenceManager::getConfig();
+    waveDurationSeconds_ = config_.waveDurationSeconds;
+    wavesPerStage_ = config_.wavesPerStage;
```
5tefan543: As also mentioned in the hpp file, I don't think it is a good idea to store values of the GameConfig in Game. The config should be directly read were needed.

src/game/game.cpp
```diff
@@ -91,36 +111,63 @@ void Game::loadFromPersistedGame(const controller::PersistedGame &persistedGame)
...
     return persistedGame;
 }
 
-bool Game::update(const controller::InputState &input, float dt)
+bool Game::isWaveTimeFinished()
 {
+    return currentWaveDuration_ >= waveDurationSeconds_;
+}
+
+controller::StateTransitionAction Game::update(const controller::InputState &input, float dt)
```
5tefan543, line 151: I think it would still be cleaner if `StateTransitionAction` is handled outside of `Game`.

Right now, the code feels more complex because state transitions are split between two places: `GameplayState` and `Game`. For example, `GameplayState` still handles transitions such as pause, while `Game` now also returns transitions such as opening the progression store or switching to the game-over state.

In my opinion, `Game` should only expose gameplay-related information, such as whether the current wave is defeated, whether the wave timer is over, whether the progression store should be opened, or whether the player is dead. `GameplayState` can then decide what this means for the application state.

So instead of `Game::update()` returning a `StateTransitionAction`, I would prefer that `GameplayState` calls `game.update()` and afterwards checks semantic game conditions like:

```cpp
StateTransitionAction GameplayState::update(const InputState &input, float dt)
{
    DebugContext &debug = DebugContext::get();
    debug.gameSession = &game.getDebugSession();

    if (input.cancelPressed) {
        debug.gameSession = nullptr;
        return controller::StateTransitionAction::PushPauseMenu;
    }

    game.update(input, dt);

    if (game.isGameOver()) {
        debug.gameSession = nullptr;
        return controller::StateTransitionAction::ReplaceCurrentWithGameOverMenu;
    }

    if (debug.active && debug.gameSession->isStoreOpenRequested) {
        debug.gameSession->isStoreOpenRequested = false;
        return controller::StateTransitionAction::PushProgressionStore;
    }

    if (game.hasCurrentWaveEnded()) {
        game.completeCurrentWave(); // e.g. add score

        if (game.shouldOpenProgressionStore()) {
            game.advanceToNextStage();
            return controller::StateTransitionAction::PushProgressionStore;
        }

        game.advanceToNextWave();
    }

    return controller::StateTransitionAction::None;
}
```
In my opinion, this feels better for several reasons:

1. If `Game` returns `StateTransitionAction`, then the game layer depends on the controller/state-machine layer. That is usually the wrong direction. The controller should depend on the game, not the game on the controller.
2. All controller/state-machine transitions stay in one place. `GameplayState` decides whether the current state should be replaced, paused, or whether another state should be pushed.
3. `Game` stays independent from controller concepts. The game logic should not need to know that opening the store means `StateTransitionAction::PushProgressionStore`.
4. Handling transitions from the Debug UI becomes easier. For example, opening the store from the `DebugUI` can be handled directly in `GameplayState`, without adding more controller-specific logic to `Game`.
5. `debug.gameSession` can be set and reset outside of `Game`. This does not really belong to the game simulation itself and is more of a controller/debug wiring concern.
6. Testing becomes clearer. `Game` tests can check gameplay conditions such as `game.hasCurrentWaveEnded()` or `game.isGameOver()`, while `GameplayState` tests can check which `StateTransitionAction` is returned.

src/game/game.cpp
```diff
@@ -131,12 +178,31 @@ void Game::processDebugSession()
...
     if (debugSession_.isStageWaveReloadRequested) {
         debugSession_.isStageWaveReloadRequested = false;
         stage_ = debugSession_.stage;
-        wave_ = debugSession_.wave;
-        initWave();
+        wave_ = debugSession_.wave - 1;
+        getNextWave();
+    } else {
+        debugSession_.stage = stage_;
+        debugSession_.wave = wave_;
```
5tefan543: This means the debugSession properties are always overridden every frame. They should only be set once after `wave_++` and `stage_++` within the respective methods `getNextWave()` `getNextStage()`.

src/game/game.cpp
```diff
@@ -131,12 +178,31 @@ void Game::processDebugSession()
...
+    }
+
     // Handle stage/wave reload request
     if (debugSession_.isStageWaveReloadRequested) {
         debugSession_.isStageWaveReloadRequested = false;
         stage_ = debugSession_.stage;
-        wave_ = debugSession_.wave;
-        initWave();
+        wave_ = debugSession_.wave - 1;
+        getNextWave();
```
5tefan543: I think this would be cleaner if `getNextWave` allowed controlling whether the wave counter should be incremented, for example:

```cpp
void getNextWave(bool updateWaveCounter = true);
```

Then the debug reload case could call `getNextWave(false)` and we would not need to temporarily set `wave_` to `debugSession_.wave - 1` just so that `getNextWave()` increments it again.

This would make the intent clearer: in the normal gameplay flow we advance to the next wave, but in the debug reload flow we want to reload the selected wave directly.

src/ui/debug_ui.cpp
```diff
@@ -74,18 +76,20 @@ void DebugUI::renderGameSession(controller::DebugContext &debug)
...
                 // derive stage from wave
-                gameSession.stage = ((gameSession.wave - 1) / 5) + 1;
+                gameSession.stage = ((gameSession.wave - 1) / wavesPerStage) + 1;
             }
 
             if (ImGui::Button("Reload Stage/Wave") && debug.gameSession) {
                 debug.gameSession->isStageWaveReloadRequested = true;
             }
 
+            ImGui::Checkbox("Pause Clock", &debug.gameSession->isClockPaused);
```
5tefan543, line 91: Is this behavior expected?

When I enable the checkbox, the timer is reset to 1 second. I would have expected the timer to simply pause, and when I enable it again, to continue from the time where it was stopped.

tests/controller/state/state_test.cpp
```diff
@@ -330,9 +331,29 @@ TEST_CASE_METHOD(TestFixture, "Gameplay state update returns correct actions")
         // ACT
         StateTransitionAction action = updateOnce();
 
+        // ASSERT
+        REQUIRE(action == StateTransitionAction::ReplaceCurrentWithGameOverMenu);
+        REQUIRE(debug.gameSession != nullptr);
+    }
```
5tefan543: I am a bit confused here.

`debug.gameSession` is a pointer to a member object inside `Game`. If we leave the `GameplayState` and the `Game` instance gets destructed, then `debug.gameSession` would still point to that object even though it no longer exists.

tests/shared/test_fixture.hpp
```diff
@@ -16,11 +19,42 @@ struct TestFixture {
...
+  },
+  "assetConfig": {
+    "playerTexturePath": "assets/characters/player.png",
+    "enemyTexturePath": "assets/characters/enemy.png",
+    "mapTexturePath": "assets/maps/test_map.png",
+    "fontPath": "assets/font/BigBlueTerm_Nerd_Font/"
+  }
+})";
+    }
+
```
5tefan543: Why not create a `GameConfig` object and use `PersistenceManager::saveConfig` for this?

Or even simpler: Why not copy the config to the test dir?

5tefan543 (changes requested):
You also have a bug. The app crashes when opening the `DebugUI` because in the current version `debug.gameSession` is not set anywhere anymore.

src/game/game.cpp
RZacchia:
```cpp
void getNextWave(int waveNumber);
```

src/game/game.cpp
5tefan543: After implementing the alternative version, I think the original comment is probably too strict.

The problem is that the wave/stage progression logic is not really controller logic. Logic such as:

```cpp
if (game.hasCurrentWaveEnded()) {
    game.completeCurrentWave();

    if (game.shouldOpenProgressionStore()) {
        game.advanceToNextStage();
        return StateTransitionAction::PushProgressionStore;
    }

    game.advanceToNextWave();
}
```

actually belongs very naturally to `Game`, because it is part of the gameplay flow and progression system.

Moving this logic into `GameplayState` made the controller more complex and forced it to orchestrate gameplay details such as wave completion, score handling, and stage progression. It also introduced synchronization issues where `GameplayState` had to manually coordinate game progression with state transitions.

So while returning `StateTransitionAction` from `Game` introduces a dependency from the game layer to the controller layer, the resulting design is actually simpler in practice because:

- the wave/stage progression stays encapsulated inside `Game`
- the gameplay flow is handled in one place
- `GameplayState` does not need to orchestrate internal game progression
- the store/game-over transitions happen exactly where the underlying gameplay events occur

include/game/game.hpp
5tefan543: Yes but it is also more work to always get the `GameConfig` from the `PersistenceManager`.

For now we init the `Game::config_` once with the config from the `PersistenceManager` in the ctor. If we later need to change the config via the `DebugUI` that is still possible via `processDebugSession()`

5tefan543 (approved):

PR #67, opened 2026-05-21
31 hitboxes and damage systems

### LINKED ISSUE #31: Hitboxes and damage systems
# TODO

### LINKED ISSUE #65: move spawning logic constants into config

RZacchia:
hitbox initialization needs follow up


include/game/persisted_game.hpp
5tefan543, line 1: Would it now not make more sense to move PersistedGame to the game namespace next to game.hpp.

include/controller/persistence/serializer.hpp
5tefan543, line 1: Do we really need the glaze mapping again? Why?

include/game/ecs/components/hitbox.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "shapes/rectangle.hpp"
+
+namespace game {
+struct HitBox {
+    Rectangle<float> rect;
+    bool isActive;
```
5tefan543: Is `isActive` even used and necessary? It is only set, I think.

src/game/systems/collision_detection_system.cpp
```diff
@@ -0,0 +1,168 @@
...
+            continue; // Skip if hitbox already exists
+        }
+
+        view::Sprite &sprite = registry.getComponent<view::Sprite>(entitiesWithHitBoxes[i]);
+        // Problem with previous idea via alpha:
+        // - Performance heavy to read pixel data for every sprite on initialization,
+        // But hitbox values can be cached
+        // Dependency to sf or other rendering library would be needed here
+        // Or Renderer will be dependend on that system both no-gos
+        // maybe make an external python script that takes pngs and gives back json hitboxes
```
5tefan543, line 69: I think this could be solved with an `AssetManager` instead of introducing an external script.

The collision system should not load or inspect image files directly, and it should also not depend on SFML or the renderer. But an abstract `AssetManager` could lazily load and cache asset metadata, including the visible content bounds of an image.

Then the game logic could ask for something like:

AssetId -> content bounds / hitbox bounds

instead of dealing with image paths or pixel data directly. The concrete implementation can decide whether these bounds are calculated from the PNG alpha data, cached, or loaded from config.

The UI/Renderer can use the same asset ids for rendering by also querying the `AssetManager`.

src/game/systems/collision_detection_system.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+void CollisionDetectionSystem::update(Registry &registry, int wave)
+{
+    if (wave != wave_) {
+        isInitialized_ = false;
+    }
+    if (!isInitialized_) {
+        initializeHitBoxes(registry);
+        wave_ = wave;
+        isInitialized_ = true;
```
5tefan543, line 146: I think this special initialization state could be avoided.

If hitbox bounds are provided through an `AssetManager` and probably also cached there, then creating missing `HitBox` components should be cheap. The collision system could simply ensure that every relevant entity has a hitbox during each update:

- if the entity has no `HitBox`, create one from the cached asset bounds
- if the entity already has a `HitBox`, update its world position from `Position`

Since hitbox positions need to be updated every frame anyway, we probably do not need a separate `isInitialized_`/`wave_` mechanism here. This would also avoid skipping collision detection on the first frame after a wave change.

src/game/systems/collision_detection_system.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    return hitBoxA.rect.intersects(hitBoxB.rect);
+}
+
+CollisionDetectionSystem::CollisionDetectionSystem() : isInitialized_(false), wave_(0) {}
+
+void CollisionDetectionSystem::initializeHitBoxes(Registry &registry)
+{
+    auto entitiesWithHitBoxes = registry.view<view::Sprite, Position>();
+    for (size_t i = 0; i < entitiesWithHitBoxes.size(); ++i) {
```
5tefan543, line 58: Use range-based loops instead:

```cpp
for (Entity entity : entitiesWithHitBoxes) {
```

src/game/systems/collision_detection_system.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (!isInitialized_) {
+        initializeHitBoxes(registry);
+        wave_ = wave;
+        isInitialized_ = true;
+        return;
+    }
+
+    const std::vector<Entity> &entitiesWithHitBoxes = registry.view<HitBox, Position>();
+
+    for (size_t i = 0; i < entitiesWithHitBoxes.size(); ++i) {
```
5tefan543, line 152: Use range-based loops instead:

```cpp
for (Entity source : entitiesWithHitBoxes) {
  ...
  for (Entity target : entitiesWithHitBoxes) {
```

**Also note that this nested loop is quite expensive. Maybe talk with Martin on how to avoid this O(N^2) lookup.** He already builds a tree for the EnemyAI, I think.

src/game/systems/damage_system.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace game {
+
+void DamageSystem::updateProjectile(Entity projectile, Registry &registry) {}
+
+void DamageSystem::update(Registry &registry, float dt)
+{
+    auto damageEntities = registry.view<Damage, DamageTag>();
+    for (size_t i = 0; i < damageEntities.size(); ++i) {
+        Entity damageEntity = damageEntities[i];
```
5tefan543: Simplify to 

```cpp
for (auto damageEntity : damageEntities) {
```

src/game/systems/input_system.cpp
```diff
@@ -1,19 +1,77 @@
...
         Velocity &velocity = registry.getComponent<Velocity>(entity);
         PlayerStats &playerStats = registry.getComponent<PlayerStats>(entity);
 
         velocity.dx = 0.0F;
         velocity.dy = 0.0F;
 
+        if (input.mouseLeftPressed) {
+            attack(registry, playerStats, entity, input);
+        }
+
```
5tefan543: Calling `attack()` mutates the registry and may add a new `Velocity` component. Since component storage is vector-based, this can reallocate the `Velocity` storage and invalidate the velocity reference below.

Therefore, either acquire the velocity reference only after `attack()`, or call `attack()` after we are done using it.

src/game/systems/movement_system.cpp
```diff
@@ -12,14 +15,26 @@ void MovementSystem::update(Registry &registry, float dt)
...
+                if (auto *projectile = std::get_if<ProjectileDamage>(&damage.params)) {
+                    projectile->distanceTraveled +=
+                        std::sqrt(velocity.dx * velocity.dx + velocity.dy * velocity.dy) * dt;
+                    if (projectile->distanceTraveled >= projectile->maxRange) {
+                        registry.destroyEntity(entity);
+                        continue;
+                    }
+                }
+            }
+        }
```
5tefan543: This logic does not really belong in `MovementSystem`.

`MovementSystem` should only be responsible for applying velocity to positions. The current projectile-specific logic introduces knowledge about `Damage`, `DamageKind::Projectile`, `ProjectileDamage`, range tracking, and entity destruction into a system that should only handle movement.

I think this should be moved into a separate system, for example `ProjectileSystem` or `ProjectileLifetimeSystem`, which updates `distanceTraveled` and destroys projectiles once they exceed their `maxRange`.

src/game/game.cpp
```diff
@@ -34,24 +35,14 @@ Game::Game(const controller::PersistedGame &persistedGame) : Game(persistedGame.
...
-        playerStats.attackSpeed = persistedGame.playerStats.attackSpeed;
-        playerStats.defense = persistedGame.playerStats.defense;
-        playerStats.moveSpeed = persistedGame.playerStats.speed;
-        playerStats.hasDash = persistedGame.playerStats.hasDash;
+        Entity player = players.front();
+        registry_.removeComponent<Position>(player);
+        registry_.addComponent<Position>(player, persistedGame.position);
+
+        registry_.removeComponent<PlayerStats>(player);
+        registry_.addComponent<PlayerStats>(player, persistedGame.playerStats);
```
5tefan543: Don't remove and re-add the components here. Instead, the `Game` should be initialized
with the correct player state from the beginning.

For example:
```cpp
Game::Game()
{
    config_ = controller::PersistenceManager::getConfig();
    initMap();
    initCamera();
    initPlayer();
    initWave(1);
}

Game::Game(const controller::PersistedGame &persistedGame)
{
    config_ = controller::PersistenceManager::getConfig();
    initMap();
    initCamera();
    initPlayer(persistedGame.position, persistedGame.playerStats);
    initWave(persistedGame.wave);
}
```

This should also fix a bug of mine in the current implementation:

`initWave()` calls `spawnEnemySystem_.update()`, which reads `PlayerStats`.
However, in the current code `initWave()` is executed before the saved `PlayerStats`
are restored. This means enemies may be spawned based on the default player stats
instead of the persisted stats.

.clang-tidy
```diff
@@ -48,9 +48,9 @@ CheckOptions:
   # Member Variables
   # =========================
 
-  # Enforce naming for ALL member variables (class + struct, all visibilities).
+  # Enforce naming for PUBLIC member variables (class + struct).
   # Example: int selectedButtonIndex;
-  - key: readability-identifier-naming.MemberCase
+  - key: readability-identifier-naming.PublicMemberCase
```
5tefan543, line 53: Why is this change needed?

5tefan543 (changes requested):

include/controller/persistence/serializer.hpp
RZacchia: Enums, throw exceptions inside glaze without mappings, but i will try if all are needed before the merge

include/game/ecs/components/hitbox.hpp
RZacchia: I know, not final

src/game/systems/collision_detection_system.cpp
RZacchia: Not final, is only here because of the way hitboxes are drawn st this point, not final

src/game/game.cpp
RZacchia: Already outdated

.clang-tidy
RZacchia: Had a weird issue with clang tidy that i could fix, i would like to keep that in for clarity

src/game/systems/input_system.cpp
RZacchia: put attack after movement changes

include/controller/persistence/serializer.hpp
RZacchia: except we could change to c++ 26 than we can remove the glaze meta again

src/controller/state/state_manager.cpp
```diff
@@ -79,6 +85,9 @@ void StateManager::applyAction(StateTransitionAction action)
         clear();
         push(ExitState::createExitState());
         break;
+    case StateTransitionAction::ReplaceCurrentWithCharacterSelection:
+        push(MenuState::createMenu(MenuType::CharacterSelection));
```
5tefan543: Use `replaceCurrent()` otherwise the `MainMenu` stays on the stack.

src/game/systems/collision_detection_system.cpp
RZacchia: lets move that into new issue pr is already big enough as is

5tefan543: Created Issue: #71

include/controller/persistence/serializer.hpp
5tefan543: Ah yes we need the mapping since you want to store the enums as string which makes sense.

Could we maybe move those mappings to custom `glaze_mappings.hpp` next to `serializer.hpp` ?

src/game/systems/collision_detection_system.cpp
5tefan543: Maybe keep this for now and improve in another sub issue?

5tefan543 (changes requested):

src/game/systems/collision_detection_system.cpp
5tefan543: Created #74

include/controller/persistence/config_game.hpp
```diff
@@ -10,10 +13,126 @@ struct WindowConfig {
...
+    std::string meleeTexturePath;
+    std::string rangedTexturePath;
     std::string enemyTexturePath;
     std::string mapTexturePath;
     std::string fontPath;
+    std::string projectilePath;
+    struct DroppableItemAssetConfig {
+        std::string id;
+        std::string spritePath;
+    };
```
5tefan543, line 26: Is there a specific reason why this struct is defined inside the `AssetConfig` struct? If not, maybe move it above the `AssetConfig`.

src/game/game.cpp
```diff
@@ -66,41 +94,93 @@ void Game::initMap()
...
     Entity camera = registry_.createEntity();
     registry_.addComponent<CameraTag>(camera, {});
     registry_.addComponent<Position>(camera, {0.0f, 0.0f});
 }
+void Game::initCamera(Position position)
+{
+    Entity camera = registry_.createEntity();
+    registry_.addComponent<CameraTag>(camera, {});
+    registry_.addComponent<Position>(camera, position);
+}
```
5tefan543: Much duplicate code. We could simplify by changing it to

```
void Game::initCamera(Position position)
{
    Entity camera = registry_.createEntity();
    registry_.addComponent<CameraTag>(camera, {});
    registry_.addComponent<Position>(camera, position);
}
```

and then always passing a position component.

src/game/game.cpp
```diff
@@ -66,41 +94,93 @@ void Game::initMap()
...
     registry_.addComponent<PlayerStats>(player, playerStats);
 
     registry_.addComponent<Position>(player, {100.0f, 100.0f});
     registry_.addComponent<Velocity>(player, {0.0f, 0.0f});
-    Animation playerAnimation = {
-        .baseTexturePath = "assets/characters/character_",
-    };
+    registry_.addComponent<Animation>(player, playerAnimation);
+    registry_.addComponent<view::Sprite>(player, {.imagePath = playerAnimation.baseTexturePath + "right_1.png"});
+}
```
5tefan543, line 134: Much duplicate code with `void Game::initPlayer(Position position, PlayerStats playerStats)`

It would be nice to simplify this.

include/controller/state/state.hpp
```diff
@@ -48,12 +48,14 @@ class GameplayState : public BaseState {
...
 
     GameplayState();
-    explicit GameplayState(const PersistedGame &persistedGame);
+    explicit GameplayState(game::CharacterType characterType);
+    explicit GameplayState(const game::PersistedGame &persistedGame);
 
   public:
     game::Game game;
 
     static std::unique_ptr<GameplayState> createNewGameplay();
```
5tefan543: Is this actually still used?

src/game/systems/collision_detection_system.cpp
```diff
@@ -0,0 +1,172 @@
+#include "game/ecs/systems/collision_detection_system.hpp"
+#include "SFML/Graphics/Image.hpp"
+#include "SFML/Graphics/RectangleShape.hpp"
+#include "SFML/Graphics/Texture.hpp"
```
5tefan543: SFML includes are not needed at this moment

src/game/systems/input_system.cpp
```diff
@@ -1,15 +1,155 @@
...
+    if (timeSinceLastAttack_ <= 1.0f / stats.attackSpeed) {
+        return; // Attack is still on cooldown
+    }
+    const view::Sprite playerSprite = registry.hasComponent<view::Sprite>(playerEntity)
+                                          ? registry.getComponent<view::Sprite>(playerEntity)
+                                          : view::Sprite{};
+    Position playerPosition = registry.getComponent<Position>(playerEntity);
+    Direction attackDirection = input.mouseGridX >= playerPosition.x ? Direction::Right : Direction::Left;
+
+    float attackDurationSec;
```
5tefan543: Should be initialized since it could be used uninitialized below.

CMakeLists.txt
```diff
@@ -34,10 +34,16 @@ add_library(roguelike_game
...
+  src/game/systems/collision_detection_system.cpp
+  src/game/systems/damage_system.cpp
 )
 
 roguelike_cxx_flags(roguelike_game)
 target_include_directories(roguelike_game PUBLIC include)
+target_link_libraries(roguelike_game
+  PUBLIC
+    SFML::Graphics
+)
```
5tefan543: This is currently not needed. Please remove this.

5tefan543 (changes requested):

include/controller/persistence/config_game.hpp
RZacchia: this will be used later for potions etc... and yeah it should be part of assettconfig

include/game/game.hpp
```diff
@@ -47,11 +49,15 @@ class Game {
     // all sprites in a deque inside Game.
     view::Text stageWaveInfo_;
 
-    explicit Game(int wave);
+    explicit Game(int wave, CharacterType characterType);
 
     void initMap();
+    void switchMap();
+    void cleanup();
     void initCamera();
```
5tefan543: Not implemented anymore. Remove.

include/controller/state/state.hpp
```diff
@@ -48,12 +48,13 @@ class GameplayState : public BaseState {
     bool loadedFromSave_ = false;
 
     GameplayState();
```
5tefan543: I don't think this ctor is still needed.

include/game/game.hpp
```diff
@@ -60,13 +66,14 @@ class Game {
 
   public:
     Game();
-    Game(const controller::PersistedGame &persistedGame);
+    explicit Game(CharacterType characterType);
+    Game(const PersistedGame &persistedGame);
```
5tefan543: `explicit` is missing for single argument ctor

5tefan543 (changes requested):

RZacchia:
closes #31 und #65

include/shapes/rectangle.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <algorithm>
+#include <cmath>
+#include <optional>
+
+template <typename T>
+struct Rectangle {
+    T x;
+    T y;
+    T width;
+    T height;
```
51th08: I would start using Vec2 for x,y (-> e.g. position or origin) and width,height (-> e.g. size) since it's already in this branch.

config/game-config.json
```diff
@@ -11,10 +11,229 @@
...
     },
     "assetConfig": {
-        "playerTexturePath": "assets/characters/character_",
-        "enemyTexturePath": "assets/characters/enemy_1_",
-        "mapTexturePath": "assets/map/map_1.png",
-        "fontPath": "assets/font/BigBlueTerm_Nerd_Font/BigBlueTerm437NerdFont-Regular.ttf"
+        "fontPath": "assets/font/BigBlueTerm_Nerd_Font/BigBlueTerm437NerdFont-Regular.ttf",
+        "meleeTexturePath": "assets/characters/melee/character_melee_",
+        "rangedTexturePath": "assets/characters/ranged/ranged_character_",
+        "enemyTexturePath": "assets/characters/enemy_blob/blob_",
```
51th08: Minor detail, why not name it [...]TexturePath**Prefix**  as it actually is used as a prefix. For me just "path" corresponds to a file that exists which is not the case here, right?

There are more of these names in this file.

include/shapes/rectangle.hpp
51th08, line 1: We now have a geometry directory with Vector and a shapes directory with rectangle. I feel like they could merged into one directory. Keeps the file structure compact.

include/shapes/rectangle.hpp
51th08: If its too much trouble we could change it in another PR since many more parts outside this PR need Vec2 updating as well.

51th08 (changes requested):

5tefan543 (approved):

51th08 (approved):

PR #69, opened 2026-05-27
Add enemy movement behaviour and location tracking table

### LINKED ISSUE #75: Init EnemyAI

51th08:
Is part of #30  but does not fully implement it (e.g. attacking is missing). However, I feel like this is a good point in time to merge this state with main.

Adds:
- Enemies move to player and avoid clumping together (repel each other a little bit)
- Location tracking table for fast entity lookup based on world coordinates.
- Wrapping of various stdout/err printing to make the terminal usable again. To enable the printing simply use `cmake -DLOG_STDOUT ...`. 

Known Issues (to be fixed in the future):
- Map is 2x bigger than grid coordinates and therefore loctation table. The table is designed to deal with that but the map should only be as big as the grid.
- Enemies can start "jittering" when surrounding the player due to little back and forth movements which changes sprites.


include/controller/persistence/serializer.hpp
RZacchia, line 1: cerr should always be printed and not only if log_stdout

src/controller/persistence/persistence_manager.cpp
RZacchia, line 1: cerr should always be printed

src/game/systems/enemy_ai.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "game/location_table.hpp"
+#include "view/sprite.hpp"
+#include <cmath>
+#include <iostream>
+
+namespace game {
+
+void EnemyAI::update(Registry &registry, LocationTable &locationTable)
+{
+    const auto players = registry.view<Velocity, PlayerTag>();
```
RZacchia: Target PlayerStat, not PlayerTag, since projectiles fired by the player also have playerTag, only PlayerStat is unique to the playablle character

src/game/systems/movement_system.cpp
RZacchia, line 1: I removed the log statement completely since i do not see any value in it anymore and really spams the console

RZacchia (changes requested):
I do not think that define is needed, if we just delete the movement update outputs (which i highly recommend) it is not needed, since the other messages are not as frequent and it really clutters the code for nearly nothing, since you can just start the game ./app > log.txt to get an even better effect.

src/game/systems/movement_system.cpp
51th08: Ok, removed #98bb01e @

51th08:
> I do not think that define is needed, if we just delete the movement update outputs (which i highly recommend) it is not needed, since the other messages are not as frequent and it really clutters the code for nearly nothing, since you can just start the game ./app > log.txt to get an even better effect.

I disagree. Redirecting simply moves the problem. Additionally, if I am interested in stdout ouput this doesn't solve anything - still have to cope with all the unnecessary log lines in log.txt.
For example, just run the roguelike_test_runner with some test failing - have fun scrolling arround and finding
the output of that failure.

In my opinion even stderr should not be required to print. That's a developer related thing. If it is meant for the user it should be done graphically. Normally you don't start GUI programs from terminal, the only place  stdout/err is directly available. You could redirect it to a system logger like systemd but thats also already far away from the concerns of a user. However, for selected logging output it would make sense to treat stderr output with a higher severity than just the stdout info of e.g. a new wave got created.

Every program I know that does some kind of logging has at least the possibility to silence it and often provides different levels of verbosity. So, if we want to have all this logging (cluttering stdout/err streams) controlling that with a simple if statement is not for nothing :)

And by the way, I would not ship a propper "release" binary with all that output, it's not good stlye! And I don't want to  go through the code, find every print statement and delete it just for a "release" version (and repeat that over and over again for every iteration...). Better start with proper logging hygiene now. If we don't want a dynamic logging option for the release build using preprocessor macros to completely remove the print statements along with any accompanying logic is the cleanest way.

include/controller/persistence/serializer.hpp
5tefan543: I agree that cerr should always be printed.

We never talked about logging. If we really want more control then I would use a logging library or implement a small logger. But do we actually need that?

I think using `#ifdef LOG_STDOUT` is a bit verbose. If we really want more control, then why not implement static `Logger::logError`, `Logger::logInfo`, `Logger::logWarn`, `Logger::logDebug` which internally use something like `#ifdef LOG_ERROR`

Or even simpler: Why not use a logging library and set the log level via `GameConfig`.

include/game/location_table.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include "game/ecs/components/position.hpp"
+#include "game/ecs/entity.hpp"
+#include "game/ecs/registry.hpp"
+#include "utils/vector.hpp"
+#include <unordered_set>
+
+namespace game {
+
+// Lookup 2D Table for finding entities based on their location in the world grid with dimensions view::gridWidth X
+// view::gridHeight.
```
5tefan543: Maybe we should start to distinguish between a `ViewGrid` when using `ViewMode::FixedToScreen` and a `WorldGrid` when using `ViewMode::FixedToWorld`.

The `LocationTable` tracks entity positions and since entities can move beyond the fixed screen coordinates of `<gridWidth, gridHeight>`, it makes sense that the table size represents the entire world coordinates.

The current clamping in `getBucketIndices()` prevents invalid bucket access, but it does not really solve the issue if the world is larger than the table. In that case, entities outside the table area are clamped into the edge buckets, which means the spatial lookup becomes less accurate and those buckets may contain too many unrelated entities.

So I think the `LocationTable` should not be based on the visible view grid size, but on the actual world/map size.

include/ui/renderer.hpp
```diff
@@ -28,6 +29,7 @@ class Renderer {
     void renderViewElement(sf::RenderWindow &window, const view::ViewElement &element);
     void renderViewElements(sf::RenderWindow &window, const std::vector<view::ViewElement> &elements);
     sf::Color toSfColor(const view::Color &color);
+    void renderDebugContext(sf::RenderWindow &window);
```
5tefan543, line 32: Maybe move to own class `DebugRenderer` used only for rendering debug logic.

include/ui/ui.hpp
```diff
@@ -24,6 +24,7 @@ class UI {
     void setSfmlView(float cameraX, float cameraY);
     sf::FloatRect getLetterboxViewport() const;
     void renderView(sf::RenderWindow &window, const view::View &view);
+    void renderDebug(sf::RenderWindow &window);
```
5tefan543: not implemented -> remove

src/game/systems/enemy_ai.cpp
```diff
@@ -0,0 +1,54 @@
...
+    if (players.size() > 0) {
+        const auto player = players[0];
```
5tefan543: Use `player.front()` instead of direct indexing.

include/game/location_table.hpp
```diff
@@ -0,0 +1,36 @@
...
+    std::vector<std::tuple<Entity, Position>> getEntitiesInRange(const Vec2<float> position, const float radius,
+                                                                 const Registry &registry);
+    const std::vector<Entity> *cgetBucket(const unsigned buckIx, const unsigned buckIy) const
+    {
+        return bucketGrid_[buckIy * numBuckets.x + buckIx].get();
+    }
+    const std::vector<std::unique_ptr<std::vector<Entity>>> &cgetGrid() { return bucketGrid_; }
+
+  private:
+    std::vector<std::unique_ptr<std::vector<Entity>>> bucketGrid_;
```
5tefan543: I think we can replace `std::vector<std::unique_ptr<std::vector<Entity>>>`
with `std::vector<std::vector<Entity>>`.

The current version allocates each bucket vector separately on the heap and stores
only pointers to those buckets in the outer vector. This adds an extra level of
indirection and makes ownership more complicated without an obvious benefit here.

Since the bucket grid is initialized once and then accessed by index, we do not
need stable heap-allocated bucket objects. The outer vector can simply store the
bucket vectors directly.

include/game/location_table.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+    LocationTable(const Vec2<unsigned> numBuckets);
+    void update(const Registry &registry);
+    std::unordered_set<Entity> getEntitiesNear(const Vec2<float> position, const float radius);
+    std::vector<std::tuple<Entity, Position>> getEntitiesInRange(const Vec2<float> position, const float radius,
+                                                                 const Registry &registry);
+    const std::vector<Entity> *cgetBucket(const unsigned buckIx, const unsigned buckIy) const
+    {
+        return bucketGrid_[buckIy * numBuckets.x + buckIx].get();
+    }
```
5tefan543: move implementation to cpp

include/game/location_table.hpp
```diff
@@ -0,0 +1,36 @@
...
+    const std::vector<std::unique_ptr<std::vector<Entity>>> &cgetGrid() { return bucketGrid_; }
+
+  private:
+    std::vector<std::unique_ptr<std::vector<Entity>>> bucketGrid_;
+    std::vector<Entity> *getBucket(const unsigned buckIx, const unsigned buckIy)
+    {
+        return bucketGrid_[buckIy * numBuckets.x + buckIx].get();
+    }
```
5tefan543: move implementation to cpp

src/game/systems/input_system.cpp
```diff
@@ -1,7 +1,9 @@
 #include "game/ecs/systems/input_system.hpp"
+#include "game/ecs/components/enemy_tag.hpp"
 #include "game/ecs/components/player_tag.hpp"
 #include "game/ecs/components/stats.hpp"
 #include "game/ecs/components/velocity.hpp"
+#include <cmath>
```
5tefan543: Is this include needed?

5tefan543 (changes requested):

src/game/systems/enemy_ai.cpp
51th08: Done 36f43c33261ef04aede923a9c7e4186d29e053d7

include/controller/persistence/serializer.hpp
51th08: Well adding another library or implementing one seems not as simple as #define for me.

Do we actually need all the non-cerr prints? I think they are responsible for most of the output.

src/game/systems/input_system.cpp
51th08: Nope. Fixed with 43f10537def4bb8937dc1c70eebcfc93b13afc2f

include/game/location_table.hpp
51th08: My intention was to inline it (which it has to be when defined in the struct afaik). It is called non-stop for every update() and the body's logic is perfectly suited for inlining.

include/game/location_table.hpp
51th08: Same as for getbucket()

src/game/systems/enemy_ai.cpp
51th08: Why? Length is checked  directly before it.

include/ui/ui.hpp
51th08: Done: da2e910ed670fd826445748f58942a948a18b713

include/game/location_table.hpp
51th08: Just use the map size then? That would be the easiest fix.

include/game/location_table.hpp
51th08: Tahnks, good catch. That's actually a remnant of some messing around. Fixed with 275c65e756e823b03c1a46c91b48c582103d62ff

include/controller/persistence/serializer.hpp
51th08: Oookey, I implemented a simple logging system: 3affecd8061d5aea436434b9b437113eecdaa78d

It allows to adjust the level and color usage with environment variables `ROUGL_LOG_LEVEL=[Silent,Error,Warning,Info,Debug]` and `ROUGL_LOG_COLOR=[On,Off]`.
This applies of course to both, the app and test_runner.

I guess that's okey for everyone and mark this resolved. If not, tell me to revisit it.

include/ui/renderer.hpp
51th08: Hmm is that really necessary? `renderDebugLocationTable()` uses `toSfFont()` which would have to be
moved into a TR shared by Renderer and DebugRenderer. It's all for just one method.

include/game/location_table.hpp
5tefan543: I meant more like refactoring the current `grid.hpp` to e.g.

```cpp
namespace view {

const float worldWidth = 3840;
const float worldHeight = 2160;

const float screenWidth = 1920;
const float screenHeight = 1080;

} // namespace view
```

and then also use `worldWidth` and `worldHeight` for initializing the map component and in the `LocationTable`.

src/game/systems/enemy_ai.cpp
5tefan543: Yes, the bounds check makes `players[0]` safe here. My point was more about readability and intent than correctness.

Since we only care about “the first player”, I think `players.front()` expresses that better than indexing with 0. I would also slightly prefer an early return to avoid nesting:

```cpp
const auto players = registry.view<Velocity, PlayerTag>();

if (players.empty()) {
    return;
}

const auto player = players.front();
```

include/game/location_table.hpp
5tefan543: Fair point. :D

My main reason was more about readability. I generally like headers where the class definition mainly shows the interface without implementation details, because I find them cleaner and easier to scan.

Would this be a good compromise?

```cpp
struct LocationTable {
    // ...

    const std::vector<Entity> &cgetBucket(unsigned buckIx, unsigned buckIy) const;
    const std::vector<std::vector<Entity>> &cgetGrid() const;

private:
    std::vector<Entity> &getBucket(unsigned buckIx, unsigned buckIy);
};
```

And then still in the header, but below the struct:

```cpp
inline const std::vector<Entity> &LocationTable::cgetBucket(
    unsigned buckIx,
    unsigned buckIy
) const
{
    return bucketGrid_[buckIy * numBuckets.x + buckIx];
}

inline const std::vector<std::vector<Entity>> &LocationTable::cgetGrid() const
{
    return bucketGrid_;
}

inline std::vector<Entity> &LocationTable::getBucket(
    unsigned buckIx,
    unsigned buckIy
)
{
    return bucketGrid_[buckIy * numBuckets.x + buckIx];
}
```

5tefan543: Also, I think it would make sense to make `LocationTable` a class instead of a struct. It has private internal state and behavior.

include/logging/log.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+#include <string>
+
+namespace logger {
+
+// Enum order is important for level hirachy comparison. Higher verbosity (higher enum value) includes all lower
```
5tefan543: typo `hierarchy`

src/game/systems/enemy_ai.cpp
```diff
@@ -0,0 +1,52 @@
...
+        for (auto enemy : registry.view<Velocity, view::Sprite, Position, EnemyTag, EnemyStats>()) {
+            // EnemyTag &enemyTag = registry.getComponent<EnemyTag>(enemy);
+            Velocity &velocity = registry.getComponent<Velocity>(enemy);
+            Position &enemyPos = registry.getComponent<Position>(enemy);
+            EnemyStats &enemyStats = registry.getComponent<EnemyStats>(enemy);
+            Vec2 posE{enemyPos.x, enemyPos.y};
+
+            // Set movement direction exactly towards player
+            Vec2 v = posP - posE;
+            auto playerDist = v.length();
```
5tefan543: unused variable

src/game/game.cpp
```diff
@@ -10,13 +11,13 @@
...
 #include "game/ecs/components/velocity.hpp"
+#include "logging/log.hpp"
 #include "view/sprite.hpp"
-#include <iostream>
 #include <view/text.hpp>
 
 namespace game {
 
-Game::Game(int wave)
+Game::Game(int wave) : locationTable_({40, 40}) // 1920/40 = 48 buckets; 1080/40 = 27 buckets
```
5tefan543: Use a named constant instead of `{40, 40}`. It makes the intent clearer and avoids unexplained magic numbers in the constructor.

src/logging/log.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace logger {
+
+struct LogSettings {
+    Level level = ERROR;
+    bool useColor = true;
+
+    LogSettings()
+    {
+        if (const char *envLevl = std::getenv("ROUGL_LOG_LEVEL")) {
```
5tefan543, line 14: I think the log level should come from `GameConfig::logLevel` instead of environment variables. Then we can also use enums and don't need to parse from strings.

What do you think of this version: #70

src/ui/renderer.cpp
```diff
@@ -134,4 +134,40 @@ void Renderer::renderElement(sf::RenderWindow &window, const view::Sprite &sprit
     }
 }
 
+void Renderer::renderDebugLocationTable(sf::RenderWindow &window)
+{
+    auto &context = controller::DebugContext::get();
+    if (!context.gameSettings.showLocationTable) {
+        return;
+    }
+    game::LocationTable &lt = context.gameSession->locationTable;
```
5tefan543, line 143: You should check whether `gameSession != nullptr` otherwise, the app crashes when the player dies while `DebugUI` is enabled.

CMakeLists.txt
```diff
@@ -25,6 +25,18 @@ find_package(ImGui-SFML CONFIG REQUIRED)
 
 find_package(glaze CONFIG REQUIRED)
 
+IF (DEFINED LOG_STDOUT)
+    add_compile_definitions(LOG_STDOUT)
+ENDIF()
```
5tefan543: Do we still need that?

5tefan543 (changes requested):

include/ui/renderer.hpp
5tefan543: Yes, I agree. For only one method it is probably not worth introducing a separate `DebugRenderer` and moving shared helpers around.

I would keep it in `Renderer` for now, but if the debug-related rendering grows, I would move it into its own class later to keep the normal rendering logic separated from debug-only visualization.

51th08: :+1:

src/logging/log.cpp
51th08: It's reasonable to provide a config option although I'm not sure if it's really necessary if we have a good default level. Currently, nothing that interacts with the logger drags anymore dependencies (e.g. Persistence chain) with it, which I think is a good thing. Since the logging level is "global" the PersitanceManager could simply set the level with your `void configure(LogSettings newSettings)` in #70 .

However, I'd like to keep the env. option. It's much more flexible and allows for quick changes ( no need to fight with config files which might not even be available)  and - importantly - applies to the test runner as well. It keeps the logging functionality independent of config/persistence.

Regarding enums I'm not sure I follow you. The log level is implemented as an enum. This little bit of parsing is necessary - I'm not aware of an alternative.

src/game/systems/enemy_ai.cpp
51th08: Done 3c0085a7328fe319d2c7dbdc8aeaa2c63578b55b

src/game/systems/enemy_ai.cpp
51th08: Done 3c0085a7328fe319d2c7dbdc8aeaa2c63578b55b

include/game/location_table.hpp
51th08: Done 66c7fa6

include/logging/log.hpp
51th08: 13c756a

src/ui/renderer.cpp
51th08: Done 6d8f710

CMakeLists.txt
51th08: Ah, nope 9489b24

src/game/game.cpp
51th08: Better? 36af600

src/logging/log.cpp
5tefan543: Yes, I see the initialization-order issue. We do not know `GameConfig::logLevel` until the config has been loaded, and loading/parsing the config may already use the logger.

But I handled that in #70 with a default logger configuration during early startup. In #70, the logger starts with `ERROR` by default, and after `GameConfig` has been loaded successfully, we call `logger::configure(...)` with the configured log level.

Also, if `GameConfig` is not available, the app will not work properly anyway, because many other settings depend on it. So for the actual application runtime, I think it makes sense to treat `GameConfig` as the main source of truth for the log level.

I personally prefer using GameConfig over relying mainly on environment variables, because env vars require the user/developer to configure something outside the application. For normal game behavior, the log level feels like part of the game configuration. Also, changing the log level is not necessarily only a developer concern. In general it can also be useful for users, for example when they want to provide more detailed logs for debugging or bug reports.

---

Regarding the enum point: I meant that if the log level comes from `GameConfig`, the application can work with `logger::LogLevel` directly after loading the config.

With Glaze, we also do not necessarily need a manual string-to-enum parser for the config, because the enum can be stored as its underlying integer value in JSON. Then `GameConfig::logLevel` is deserialized directly as the enum value.

If we want the JSON to contain readable strings like `"Debug"` or `"Info"` instead of integers, then we would need a Glaze enum mapping. But that mapping would belong to the serialization/config layer, not to the logger itself.

51th08: ` JSON to contain readable strings like "Debug" or "Info" instead of integers` I'd prefer that. It's annoying to look up what number corresponds to which level and gets more and more confusing if more intermediate levels are added later.

`Also, if GameConfig is not available, the app will not work properly anyway` My primary concern is to modify the log level during development and testing, not for the finished game. For development/testing/debugging you can have portions that do log and are (maybe temporarily) set up to work without GameConfig.

That the env. variable is not "really" a user interface is ok (actually, only for windows and mac? since it's normal for linux) but what does specifically stand against keeping the env interface (with higher/overwrite priority than the config) for us developers? I don't mind if you want to add a config option.

I just want a quick and easy way to adjust the log level for both the game and test runner to isolate the output I require. A config is not suitable for that.

In our case logging is actually only a debug tool. Anything important to the user should be done graphically (in this case stdout/err is as much an interface as env. variables). As soon as the user wants to provide debugging logs for bug reporting he is outside the intended user experience. It is therefore even fitting to configure something "outside the application".

include/game/location_table.hpp
51th08: Do we want to add this second `world` definition? Some dependencies would have to be adjusted:
- The map sprite size has to be forced to be this size (or shall we allow bigger sizes?)
- The world hitbox has to conform to the `world` size. Not sure what @RZacchia planned for the world hitbox. At least for enemiy/player sprites he wanted to automatically fit it to a non-transparent pixel-based sub-recangle, afaik.

As an alternative, we could make the overall table size into a constructor argument. The game
could then initiate the map and pass its size to the table's constructor when initiating it. I might even like this approach more since the table gets rid of this last "environment" state and becomes a completely modular unit.

5tefan543: I mean the world coordinates are actually the map coordinates. So for me it would make sense that the map sprite size conforms to the world size, or at least that the world/map size is the source of truth for systems that work in world coordinates.

But I agree with your ctor idea. It would probably be best if the `LocationTable` itself does not define or fetch the table size from some global/view state. Instead, the size should be decided when creating it.

src/logging/log.cpp
5tefan543: Okay, I agree that readable strings in the JSON are better than storing the enum as an integer.

Keeping the env variable as a developer/test override is fine for me.

include/game/location_table.hpp
51th08: What to you think of 32b1d2a?

src/logging/log.cpp
51th08: Alright. Do you want to add the config option?
Should we merge this branch before that?

include/game/location_table.hpp
5tefan543: I don't think creating a dummy `LocationTable` with

```cpp
Game::Game(int wave) : locationTable_({0, 0}, {0, 0})
```

and then manually destroying and reconstructing it with `std::destroy_at` / `std::construct_at` is a good way to solve this. It introduces manual lifetime management for a normal member object, even though we can construct it correctly once.

What do you think about moving the required values into `GameConfig` and initializing the `LocationTable` directly from there?

```cpp
class Game {
  private:
    Registry registry_;
    controller::GameConfig config_;
    LocationTable locationTable_;
    GameDebugSession debugSession_{registry_, locationTable_};
    ...
};
```

```cpp
struct GameConfig {
    Vec2<unsigned> locTabNumBuckets = {20, 20};
    Vec2<float> mapSize = {1920.0f * 2.0f, 1080.0f * 2.0f};
    ...
};
```

```cpp
Game::Game(int wave)
    : config_(controller::PersistenceManager::getConfig()),
      locationTable_(config_.locTabNumBuckets, config_.mapSize)
{
    initMap();
    initCamera();
    initPlayer();
    initWave(wave);
}
```

As far as I understand, class members are initialized in the order in which they are declared in the class, not in the order in which they appear in the constructor initializer list. That's why I changed the member order so that `config_` is initialized before `locationTable_`.

src/logging/log.cpp
5tefan543: Yes I can add that to the logger but maybe in another Issue?

5tefan543: #72

5tefan543 (changes requested):
Don't forget the tests

51th08:
> Don't forget the tests

Vec2 tests or did you mean something else?

include/game/location_table.hpp
51th08: True, it's kind of a hack. Better: 2968c6f ?

5tefan543: Looks good now! But can you also update the `game-config.json` to include the new properties and maybe remove the default values from the `GameConfig` struct.

include/game/game.hpp
```diff
@@ -8,25 +8,30 @@
...
+#include "game/location_table.hpp"
 #include "view/view.hpp"
 
 namespace game {
 
 class Game {
   private:
     Registry registry_;
-    GameDebugSession debugSession_{registry_};
+    const Vec2<unsigned> locTabNumBuckets = {20, 20};
```
5tefan543: This should no longer be necessary, since it is now stored in `GameConfig`, right?

include/controller/persistence/config_game.hpp
```diff
@@ -25,6 +26,8 @@ struct GameConfig {
     int maxEnemyCount;
     WindowConfig windowConfig;
     AssetConfig assetConfig;
+    Vec2<unsigned> locTabNumBuckets = {20, 20};
+    Vec2<float> mapSize = {1920.0f * 2.0f, 1080.0f * 2.0f};
```
5tefan543: Maybe remove default values and add properties to `game-config.json` file. We don't use default values for any other property.

src/game/game.cpp
```diff
@@ -10,13 +11,15 @@
...
-#include <iostream>
+#include <format>
 #include <view/text.hpp>
 
 namespace game {
 
 Game::Game(int wave)
+    : config_(controller::PersistenceManager::getConfig()), locationTable_(config_.locTabNumBuckets, config_.mapSize)
 {
     config_ = controller::PersistenceManager::getConfig();
```
5tefan543: I think this is redundant now, since `config_` is already initialized in the constructor initializer list.

5tefan543 (changes requested):
Regarding tests, I meant that we could also consider adding tests for the logger and the enemy AI system, since the [Codecov PR](https://app.codecov.io/gh/5tefan543/uibk_cpp_project/pull/69?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Stefan+Rupert+Walter+Wagner#diff-c3JjL2xvZ2dpbmcvbG9nLmNwcA==) coverage report shows 27 missed lines for the logger and 6 missed lines for the enemy AI system.

src/game/game.cpp
51th08: Done 4df05ad

include/game/game.hpp
51th08: Yes, removed 4df05ad

include/controller/persistence/config_game.hpp
51th08: Done fd965a5

include/game/location_table.hpp
51th08: Done fd965a5df62283720a958d22b1c68f07f3ededd3

src/game/systems/enemy_ai_system.cpp
```diff
@@ -0,0 +1,77 @@
...
+        Vec2 posE{px, py};
+
+        // Set movement direction exactly towards player
+        Vec2 v = posP - posE;
+
+        // TODO: add player attack:
+        // if (v.length() < 30) {attack_player();}
+
+        // Prevent shooting over target (player) position
+        if (v.length() < 5) {
```
5tefan543: Could you add a named constant for 5 at the top of the file? That would make the meaning of this threshold clearer and avoid the magic number here.

src/game/systems/enemy_ai_system.cpp
```diff
@@ -0,0 +1,77 @@
...
+            vx = 0; // TODO: into Vec2
+            vy = 0; // TODO: into Vec2
+            continue;
+        }
+
+        // Calc. repelling force between enemies
+        auto enemiesInRange = locationTable.getEntitiesInRange(posE, 50, registry);
```
5tefan543: Could we replace 50 with a named constant?

src/game/systems/enemy_ai_system.cpp
```diff
@@ -0,0 +1,77 @@
...
+        Vec2<float> repelOffset = {0, 0};
+        for (auto [e, position] : enemiesInRange) {
+            if (e == enemy) {
+                continue;
+            }
+            const auto p = Vec2{position.x, position.y}; // TODO: into Vec2
+            const auto pToOther = (posE - p);
+            repelOffset += pToOther / (std::pow(pToOther.length(), 1.5)); // increase repelling with proximity
```
5tefan543: Could we replace 1.5 with a named constant?

src/game/systems/enemy_ai_system.cpp
```diff
@@ -0,0 +1,77 @@
...
+        v *= enemyStats.moveSpeed;
+
+        // Enforce enemy speed limit if repelling force would boost it over max
+        auto l = std::min(v.length(), enemyStats.moveSpeed);
+
+        // Prevents jittery movement in enemy heaps by
+        // - decreasing speed inverse propotional to max speed
+        // - and stopping movement below a certaing percentage
+        l *= std::pow(l / enemyStats.moveSpeed, 2);
+        if (l / enemyStats.moveSpeed < 0.02) {
```
5tefan543: Could we replace 0.02 with a named constant?

src/game/systems/enemy_ai_system.cpp
5tefan543: Should we guard against `pToOther.length()` being zero or very small here? Two enemies could still occupy the same position, which would make the repel calculation unstable.

src/game/systems/enemy_ai_system.cpp
```diff
@@ -0,0 +1,77 @@
...
+        v += repelOffset;
+        v *= enemyStats.moveSpeed;
+
+        // Enforce enemy speed limit if repelling force would boost it over max
+        auto l = std::min(v.length(), enemyStats.moveSpeed);
+
+        // Prevents jittery movement in enemy heaps by
+        // - decreasing speed inverse propotional to max speed
+        // - and stopping movement below a certaing percentage
+        l *= std::pow(l / enemyStats.moveSpeed, 2);
```
5tefan543, line 75: Should we handle the case where `enemyStats.moveSpeed` is near 0 before dividing by it? Even if it should not happen, a guard would make this more robust.

src/game/systems/enemy_ai_system.cpp
```diff
@@ -0,0 +1,77 @@
...
+        if (l / enemyStats.moveSpeed < 0.02) {
+            vx = 0; // TODO: into Vec2
+            vy = 0; // TODO: into Vec2
+            continue;
+        }
+        v.setLenght(l);
```
5tefan543: Small typo: should this be `setLength` instead of `setLenght`?

tests/game/location_table_test.cpp
```diff
@@ -0,0 +1,251 @@
...
+//         for (size_t i = 0; i < ltSpy.getNumBuckW(); i++) {
+//             std::cout << "(" << l << ", " << i << "): ";
+//             for (auto e : *ltSpy.getBucket(i, l)) {
+//                 std::cout << e << " ";
+//             }
+//             std::cout << std::endl;
+//         }
+//     }
+//     fflush(stdout);
+// }
```
5tefan543: Are these old / outdated comments? Would it make sense to remove them?

tests/game/location_table_test.cpp
```diff
@@ -0,0 +1,251 @@
...
+#include "game/ecs/components/velocity.hpp"
+#include "game/location_table.hpp"
+#include "view/grid.hpp"
+
+#include "view/sprite.hpp"
+#include <SFML/Graphics/Rect.hpp>
+
+#include <algorithm>
+#include <catch2/catch_test_macros.hpp>
+#include <cmath>
```
5tefan543, line 10: Could you check which includes are needed. I don't think `#include <SFML/Graphics/Rect.hpp>` is needed.

5tefan543 (changes requested):

src/game/systems/enemy_ai_system.cpp
51th08: Oh do you mean unstable -> `repelOffset +=  0 / 0` ?

5tefan543: Yes exactly. Or also `some_number / 0` or `some_number / 0.00001`

src/game/systems/enemy_ai_system.cpp
51th08: b44b230

src/game/systems/enemy_ai_system.cpp
51th08: b44b230

src/game/systems/enemy_ai_system.cpp
51th08: b44b230

src/game/systems/enemy_ai_system.cpp
51th08: b44b230

src/game/systems/enemy_ai_system.cpp
51th08: b44b230

src/game/systems/enemy_ai_system.cpp
51th08: b44b230

src/game/systems/enemy_ai_system.cpp
51th08: b44b230

tests/game/location_table_test.cpp
51th08: b44b230

tests/game/location_table_test.cpp
51th08: b44b230

5tefan543 (approved):

RZacchia (approved):

PR #70, opened 2026-06-02
improve logging: load logLevel from GameConfig

5tefan543:


5tefan543:
already solved in #79

PR #87, opened 2026-06-25
Refactor Vec2 to SFML::Vector2 conversion & cleanup

51th08:
914f6827c2e5aad7052814d220f89e53a4f5f7c8


include/geometry/vector.hpp
5tefan543: remove not needed import

also from CMakeLists.txt (the config fix)

5tefan543 (changes requested):

include/geometry/vector.hpp
51th08: yikes, thanks

5tefan543 (approved):

PR #93, opened 2026-06-25
Add dynamic frametime. Rename dt values to dtSec to make intentions c…

### LINKED ISSUE #92: Make frametime dt dynamic
Currently frame time was hard coded to 1/60.
The intention was to make it dynamic, based on how fast the game loop updates.

51th08:
…learer

### DIFF
--- a/apps/main.cpp
+++ b/apps/main.cpp
@@ -1,8 +1,6 @@
 #include "controller/controller.hpp"
-#include "game/ecs/components/animation.hpp"
-#include "game/ecs/components/stats.hpp"
+#include "controller/timing.hpp"
 #include "logging/log.hpp"
-#include "ui/frametime.hpp"
 #include "ui/ui.hpp"
 #include "view/view.hpp"
 #include <atomic>
@@ -33,16 +31,16 @@ int main()
         controller::Controller controller;
         ui::UI ui;
 
-        std::chrono::high_resolution_clock::time_point endFrameTime;
-        std::chrono::high_resolution_clock::time_point startFrameTime = std::chrono::high_resolution_clock::now();
+        std::chrono::steady_clock::time_point endFrameTime;
+        std::chrono::steady_clock::time_point startFrameTime = std::chrono::steady_clock::now();
 
         while (ui.isOpen() && !shutdownRequested) {
             const controller::InputState &input = ui.pollInput();
 
-            endFrameTime = std::chrono::high_resolution_clock::now();
-            const ui::frametimeDelta dt = endFrameTime - startFrameTime;
+            endFrameTime = std::chrono::steady_clock::now();
+            const controller::timeDelta dt = endFrameTime - startFrameTime;
             startFrameTime = endFrameTime;
-            controller.update(input, std::chrono::duration<double, std::milli>(dt).count() / (double)1e3);
+            controller.update(input, dt);
 
             controller::BaseState &currentState = controller.getCurrentState();
             if (typeid(currentState) == typeid(controller::ExitState)) {
--- a/apps/main.cpp
+++ b/apps/main.cpp
@@ -2,6 +2,7 @@
 #include "game/ecs/components/animation.hpp"
 #include "game/ecs/components/stats.hpp"
 #include "logging/log.hpp"
+#include "ui/frametime.hpp"
 #include "ui/ui.hpp"
 #include "view/view.hpp"
 #include <atomic>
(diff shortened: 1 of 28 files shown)


51th08:
I deem this to be ready. If we want:
- to turn off SFML's frame limit
- to add a GUI button to set the target framerate
then this could be done in a new Issue.

apps/main.cpp
```diff
@@ -32,18 +33,23 @@ int main()
...
+        std::chrono::high_resolution_clock::time_point startFrameTime = std::chrono::high_resolution_clock::now();
 
         while (ui.isOpen() && !shutdownRequested) {
             const controller::InputState &input = ui.pollInput();
 
-            controller.update(input, fixedDt);
+            endFrameTime = std::chrono::high_resolution_clock::now();
+            const ui::frametimeDelta dt = endFrameTime - startFrameTime;
+            startFrameTime = endFrameTime;
+            controller.update(input, std::chrono::duration<double, std::milli>(dt).count() / (double)1e3);
```
5tefan543: We could also pass a duration directly like

`void update(const InputState &input, std::chrono::duration<double> dt);`

or with an alias

`void update(const InputState &input, controller::DeltaTime dt);`

that would be even nicer I think, because the code itself can then decide what time units to work with.

And also Thoman mentioned this once during a lecture, I think.

apps/main.cpp
```diff
@@ -32,18 +33,23 @@ int main()
         controller::Controller controller;
         ui::UI ui;
 
-        const float fixedDt = 1.0f / 60.0f; // Fixed time step for updates
+        std::chrono::high_resolution_clock::time_point endFrameTime;
```
5tefan543: Is there a specific reason for using `high_resolution_clock` here?

For frame timing, we probably care less about maximum precision and more about measuring elapsed time reliably. `high_resolution_clock` is not guaranteed to be steady and may be an alias for `system_clock` depending on the implementation. That means it could theoretically be affected by system time adjustments.

Would `std::chrono::steady_clock` be a better fit here? It is monotonic, so it is guaranteed not to jump backwards, which makes it safer for calculating frame deltas.

https://www.sandordargo.com/blog/2025/12/10/clocks-part-4-high_resolution_clock#conclusion:~:text=Conclusion,clocks.%20Stay%20tuned.

5tefan543 (changes requested):

apps/main.cpp
51th08: I get your point. Do you want to change every last `float dtSec` parameter? Then, e.g., every system updated in Game will make the same conversion to seconds.

apps/main.cpp
51th08: Did not know that :+1:  Will change it to steady_clock

apps/main.cpp
51th08: Like this [283519d](https://github.com/5tefan543/uibk_cpp_project/pull/93/commits/283519d1f88cf38a56a2e3cf8462203d697d1957) ? oder pass the duration further down? We do not really need chono::duration (also in the future) and the parameter explicitly tells us to pass seconds.

include/controller/controller.hpp
```diff
@@ -13,7 +13,7 @@ class Controller {
   public:
     Controller();
     ~Controller();
-    void update(const InputState &input, float dtSec);
+    void update(const InputState &input, const controller::timeDelta &dtSec);
```
5tefan543: Could we maybe use `dt` and `dtSec` consistently here?

I would use `dt` when the value is a `controller::timeDelta`, and `dtSec` when it represents seconds as a float.

include/ui/ui.hpp
```diff
@@ -31,7 +31,7 @@ class UI {
     bool isOpen() const;
     const controller::InputState &pollInput();
 
-    void render(const view::View &view, const frametimeDelta &dtSec);
+    void render(const view::View &view, const controller::timeDelta &dtSec);
```
5tefan543: Could we maybe use `dt` and `dtSec` consistently here?

I would use `dt` when the value is a `controller::timeDelta`, and `dtSec` when it represents seconds as a float.

src/ui/debug_ui.cpp
```diff
@@ -17,7 +16,7 @@ DebugUI::~DebugUI()
     logger::log(logger::DEBUG, "DebugUI destructed");
 }
 
-void DebugUI::render(const controller::InputState &input, const frametimeDelta &dtSec)
+void DebugUI::render(const controller::InputState &input, const controller::timeDelta &dtSec)
```
5tefan543: Could we maybe use `dt` and `dtSec` consistently here?

I would use `dt` when the value is a `controller::timeDelta`, and `dtSec` when it represents seconds as a float.

include/ui/debug_ui.hpp
```diff
@@ -38,7 +38,7 @@ class DebugUI {
     DebugUI();
     ~DebugUI();
 
-    void render(const controller::InputState &input, const frametimeDelta &dtSec);
+    void render(const controller::InputState &input, const controller::timeDelta &dtSec);
```
5tefan543: Could we maybe use `dt` and `dtSec` consistently here?

I would use `dt` when the value is a `controller::timeDelta`, and `dtSec` when it represents seconds as a float.

apps/main.cpp
5tefan543: Yes, I agree. Changing every `float dtSec` parameter would be kind of overkill.

I think the current approach is a good compromise: pass `controller::DeltaTime` only to the components where it makes sense, and let each component decide internally which time unit it wants to work with.

5tefan543 (changes requested):

include/ui/debug_ui.hpp
51th08: 3e0fdd1

src/ui/debug_ui.cpp
51th08: 3e0fdd1

include/ui/ui.hpp
51th08: 3e0fdd1

include/controller/controller.hpp
51th08: 3e0fdd1

5tefan543 (approved):

PR #94, opened 2026-06-25
63 init progression store

### LINKED ISSUE #63: Init Progression Store

5tefan543:


src/game/store/progression_store_helper.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+std::vector<float> normalizeWeights(const std::vector<float> &weights)
+{
+    std::vector<float> probabilities = weights;
+
+    const float sum = std::accumulate(probabilities.begin(), probabilities.end(), 0.0f);
+
+    if (probabilities.empty()) {
+        return probabilities;
+    }
```
RZacchia: early return if empty would be better before summing up and copying vector

src/controller/state/state.cpp
```diff
@@ -361,92 +361,45 @@ const view::View &GameplayState::getView()
     return view_;
 }
 
-ProgressionStoreState::ProgressionStoreState()
+ProgressionStoreState::ProgressionStoreState(game::Game &game) : store_(game)
```
RZacchia, line 364: why not const game?

src/game/store/progression_store.cpp
```diff
@@ -0,0 +1,682 @@
...
+    addStatsRow("Dash", [this]() { return std::string(playerStats_.hasDash ? "Yes" : "No"); });
+
+    return statsCard;
+}
+
+view::Card &ProgressionStore::createStoreItemsCard()
+{
+    view::Card &itemsCard = cards_.emplace_back(view::Card());
+    itemsCard.rect = geometry::Rectangle<float>{.position = {itemsX, contentY}, .size = {itemsW, contentH}};
+    itemsCard.backgroundColor = view::Color{22, 22, 22};
```
RZacchia: I would prefer if we have color codes stored somewhere that we can use like
```cpp static const view::Color MEDIUM_GREY = {22,22,22};``` But I would not consider this really necessary now since we have so few Color uses

**Review by RZacchia — CHANGES_REQUESTED** [review]

File: src/game/store/progression_store_helper.cpp (hunk 1 above)
Comment (5tefan543) [reply src/game/store/progression_store_helper.cpp]: I agree -> [727e914](https://github.com/5tefan543/uibk_cpp_project/pull/94/commits/727e9142971dfd4b9801b13f4dce480f053d5645)

File: src/controller/state/state.cpp (hunk 2 above)
Comment (5tefan543) [reply src/controller/state/state.cpp:364]: Because we actively need to change the game in the store.

For example, we call `game_.setShouldOpenStore(false);` when we click on the `NextStage` button.

```cpp
void Game::setShouldOpenStore(bool shouldOpenStore)
{
    shouldOpenStore_ = shouldOpenStore;
}
```

src/game/store/progression_store.cpp
5tefan543: Yes, I like this idea. -> [5d73687](https://github.com/5tefan543/uibk_cpp_project/pull/94/commits/5d73687c6460230c7b158c96a72e6583341cada5)

src/controller/state/state.cpp
RZacchia: Ah i missed this.

RZacchia (approved):

PR #95, opened 2026-07-01
Use location table for collision lookup for \!EnemyTag->EnemyTag

### LINKED ISSUE #74: Improve performance of CollisionDetectionSystem
> Use range-based loops instead:
> 
> ```cpp
> for (Entity source : entitiesWithHitBoxes) {
>   ...
>   for (Entity target : entitiesWithHitBoxes) {
> ```
> 
> **Also note that this nested loop is quite expensive. Maybe talk with Martin on how to avoid this O(N^2) lookup.** He already builds a tree for the EnemyAI, I think. 

 _Originally posted by @5tefan543 in [#67](https://github.com/5tefan543/uibk_cpp_project/pull/67/changes#r3298804932)_

51th08:
Remove collision lookup for EnemyTag->EnemyTag which has no effect

### DIFF
--- a/include/game/ecs/registry.hpp
+++ b/include/game/ecs/registry.hpp
@@ -149,13 +149,13 @@ class Registry {
     class HasAllOf {};
 
     template <typename... Components>
-    class HasOneOrMore {};
+    class HasAnyOf {};
 
     template <typename... Components>
     class HasNoneOf {};
 
     template <typename... InclAnd, typename... InclOr, typename... Excl>
-    std::vector<Entity> view(HasAllOf<InclAnd...>, HasOneOrMore<InclOr...>, HasNoneOf<Excl...>) const
+    std::vector<Entity> view(HasAllOf<InclAnd...>, HasAnyOf<InclOr...>, HasNoneOf<Excl...>) const
     {
         std::vector<Entity> matchingEntities;
 
--- a/include/game/ecs/registry.hpp
+++ b/include/game/ecs/registry.hpp
@@ -144,6 +144,26 @@ class Registry {
 
         return matchingEntities;
     }
+
+    template <typename... Components>
+    class FilterIncl {};
+
+    template <typename... Components>
+    class FilterExcl {};
+
+    template <typename... Incl, typename... Excl>
+    std::vector<Entity> view(FilterIncl<Incl...>, FilterExcl<Excl...>) const
+    {
+        std::vector<Entity> matchingEntities;
+
+        for (Entity entity : aliveEntities_) {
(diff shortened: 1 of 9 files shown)


51th08:
For the issue of missing (projectile) hits due to their speed being fast in correlation to the frame update duration to skip over hitboxes I would open another issue. It was not really the goal with this one. You ok with that?

include/game/ecs/registry.hpp
```diff
@@ -146,19 +146,30 @@ class Registry {
     }
 
     template <typename... Components>
-    class FilterIncl {};
+    class HasAllOf {};
 
     template <typename... Components>
-    class FilterExcl {};
+    class HasOneOrMore {};
```
5tefan543: Could we rename this to `HasAnyOf` ?

src/game/systems/collision_detection_system.cpp
```diff
@@ -107,47 +107,33 @@ void CollisionDetectionSystem::enforceMapBound(const Entity &entity, Registry &r
...
+    // === Check/Activate Collision/Damage Player's -> Enemies'
+    const std::vector<Entity> nonEnemies = registry.view(
+        Registry::HasAllOf<HitBox, Position>(), Registry::HasOneOrMore<>(), Registry::HasNoneOf<MapTag, EnemyTag>());
+
+    for (const Entity nonEnemy : nonEnemies) {
+        enforceMapBound(nonEnemy, registry);
+
+        // TODO: simply leave radius hardcoded? Currently all enemy sprites are of size 128x128
+        const auto p = registry.getComponent<Position>(nonEnemy).p;
+        for (const auto enemyNear : locationTable.getEntitiesNear(p, 130)) {
```
5tefan543: Could you refactor the `130` to a `constexpr` ?

5tefan543 (changes requested):

5tefan543:
> For the issue of missing (projectile) hits due to their speed being fast in correlation to the frame update duration to skip over hitboxes I would open another issue. It was not really the goal with this one. You ok with that?

Fine for me. Just create the issue and append it to your comment.

src/game/systems/collision_detection_system.cpp
51th08: 8a2b569

include/game/ecs/registry.hpp
51th08: 8a2b569

51th08:
> For the issue of missing (projectile) hits due to their speed being fast in correlation to the frame update duration to skip over hitboxes I would open another issue. It was not really the goal with this one. You ok with that?

=> https://github.com/5tefan543/uibk_cpp_project/issues/96

5tefan543 (approved):

PR #97, opened 2026-07-02
Special Ability B: Ranged Player 

### LINKED ISSUE #27: Special Ability B: Todo
# TODO

5tefan543:


RZacchia (approved):

include/game/store/progression_store.hpp
```diff
@@ -4,6 +4,7 @@
 #include "game/game.hpp"
 #include "view/view.hpp"
 #include <deque>
+#include <functional>
```
51th08, line 7: Why add this include?

51th08 (changes requested):

include/game/store/progression_store.hpp
5tefan543: `std::function<std::string()> getText;` requires `<functional>`.

I noticed this during a refactor of the audio controller, where the missing include caused a compile error. It may currently work because `<functional>` is pulled in indirectly through another header, but adding the include here makes the dependency explicit and prevents this code from breaking if that indirect include is removed later.

51th08: I see: 👍

51th08 (approved):

PR #104, opened 2026-07-05
96 some projectile collisions do not register

### LINKED ISSUE #96: Some projectile collisions do not register
Some projectile collisions do not register due to their speed being fast enough in correlation to the frame update duration to skip over hitboxes.

https://github.com/5tefan543/uibk_cpp_project/issues/95#issuecomment-4866180009_

51th08:


src/game/systems/collision_detection_system.cpp
```diff
@@ -104,22 +145,37 @@ void CollisionDetectionSystem::enforceMapBound(const Entity &entity, Registry &r
...
-        // TODO: simply leave radius hardcoded? Currently all enemy sprites are of size 128x128
         const auto p = registry.getComponent<Position>(nonEnemy).p;
-        for (const auto enemyNear : locationTable.getEntitiesNear(p, locTableLookupRadius)) {
-            if (checkCollision(nonEnemy, enemyNear, registry)) {
-                activateDamage(nonEnemy, enemyNear, registry);
+
+        DamageKind dmgKind;
+        if (registry.hasComponent<PlayerAttackTag>(nonEnemy) && registry.hasComponent<Damage>(nonEnemy)
+            && ((dmgKind = registry.getComponent<Damage>(nonEnemy).kind) == DamageKind::Projectile
+                || dmgKind == DamageKind::Unicorn)) {
```
5tefan543, line 153: Could we maybe split this `if` into smaller checks? It is a bit hard to read at the moment.

In my opinion, it would be clearer to structure it like this:

1. Check whether the entity has `PlayerAttackTag` and `Damage`.
2. Get the `Damage` component.
3. Check whether `damage.kind` is `DamageKind::Projectile` or `DamageKind::Unicorn`.

src/game/systems/collision_detection_system.cpp
```diff
@@ -104,22 +145,37 @@ void CollisionDetectionSystem::enforceMapBound(const Entity &entity, Registry &r
...
+            // Projectile collision handled extra because they can, with a combination of high velocity and high dtSec,
+            // skip over entire enemy hitboxes without registering a collision.
+            const auto v = registry.getComponent<Velocity>(nonEnemy).v;
+            const auto radius = locTableLookupDistance + (v * dtSec).length();
+            for (const auto enemyNear : locationTable.getEntitiesNear(p, radius)) {
+                if (checkProjectileCollision(nonEnemy, enemyNear, registry, dtSec)) {
```
5tefan543, line 159: Could we maybe rename `checkProjectileCollision`? It now also handles the unicorn attack case, so the current name feels a bit too specific.

5tefan543 (changes requested):

src/game/systems/collision_detection_system.cpp
51th08: My intention with this long line is to retain the ability of a boolean short-circuit without nesting every check with an if-statement. I thinks is readable enough and can be excused for performance gains :)

src/game/systems/collision_detection_system.cpp
51th08: Isn't the unicorn some kind of projectile? I'm open for naming suggestions. Something like checkFastMovingEntitiesThatDoDamage ?

RZacchia (dismissed):

src/game/systems/collision_detection_system.cpp
5tefan543: Ok

5tefan543 (approved):

